PR #6, opened 2020-09-22
Expose private func readAccessFileToTls as public ReadAccessFileToTls 

zawachte-msft:
Expose private func readAccessFileToTls as public ReadAccessFileToTls. 

It is a useful function to be exported from the moc auth pkg


JocelynBerrendonner (approved):
LGTM

PR #10, opened 2020-09-29
Cleanup protobuf generation.

cwize1:
1. Create protobuf generation script that is callable in both Azure Pipelines
and on local dev boxes. This mostly follows the existing Azure Pipelines
script but with a little bit of cleanup and optimizations.

2. Ensure only a single version of the protobuf Go libraries are being used
(by running the script).


rpc/nodeagent/network/virtualnetwork.pb.go
```diff
@@ -7,12 +7,12 @@ import (
 	context "context"
 	fmt "fmt"
 	proto "github.com/golang/protobuf/proto"
+	empty "github.com/golang/protobuf/ptypes/empty"
+	wrappers "github.com/golang/protobuf/ptypes/wrappers"
 	common "github.com/microsoft/moc/rpc/common"
 	grpc "google.golang.org/grpc"
 	codes "google.golang.org/grpc/codes"
 	status "google.golang.org/grpc/status"
-	emptypb "google.golang.org/protobuf/types/known/emptypb"
```
cwize1, line 14: FYI: It seems like a previous change accidentally used the wrong Go runtime libraries version when generating these files. This change ensures this file is consistent with the other `*.pb.go` files.

rpc/gen_proto.sh
```diff
@@ -23,50 +23,50 @@ protoc -I common common/security.proto --go_out=plugins=grpc:$GOPATH/src
 
 Module="admin"
 echo "Generating $Agent/$Module protoc"
-protoc -I $Agent/$Module/credentialmonitor -I ./common -I $GOPATH/src/github.com/protocolbuffers/protobuf/src $Agent/$Module/credentialmonitor/credentialmonitor.proto  --go_out=plugins=grpc:$GOPATH/src
```
cwize1, line 26: Through experimentation I discovered that `protoc` will check the `PATH` environment variable for any dependencies it needs. Taking advantage of this means we can remove the dependency on having these files live in a specific directory.

madhanrm (approved):

PR #12, opened 2020-10-23
Create controlplane API to enable HA CloudAgent without FailOver Cluster

bfjelds:
This PR is in support of https://github.com/microsoft/wssdcloudagent/pull/349

[rebased onto v0.10.5 on 12/4]


madhanrm (approved):

rpc/cloudagent/cloud/controlplane/controlplane.proto
```diff
@@ -0,0 +1,36 @@
...
+	Operation OperationType = 2;
+}
+
+message ControlPlaneResponse {
+	repeated ControlPlane ControlPlanes = 1;
+	google.protobuf.BoolValue Result = 2;
+	string  Error = 3;
+}
+
+message ControlPlane {
```
jiria, line 22: Does this correspond to a running cloud agent instance? Can we add some comments here?

rpc/cloudagent/cloud/controlplane/controlplane.proto
```diff
@@ -0,0 +1,36 @@
...
+	string id = 1;
+	string name = 2;
+	string locationName = 3;
+	string fqdn = 4;
+	int32 port = 5;
+	string certificate = 6;
+	int32  authorizerPort = 7;
```
jiria, line 29: What is this one?

bfjelds: cert and authPort were planned to provide analagous behavior to node, but are not needed and have been removed.

pkg/net/net.go
```diff
@@ -111,3 +112,45 @@ func PrefixesOverlap(cidr1 net.IPNet, cidr2 net.IPNet) bool {
 	}
 	return false
 }
+
+func GetNetworkInterface() (string, error) {
```
JocelynBerrendonner, line 116: Out of curiosity: how is this used? If multiple physical interfaces are connected to the host, GetNetworkInterface will return one of the random interfaces that match the constraint. But from one run to another on the same multi-nic host, the interface returned by this function will differ. Is this acceptable? If not, did you consider using the approach used in GetIPAddress() (which is designed to always return the public facing IP - which should be consistent over time)?
Also, as a nitpick, I think we should name this function accordingly (GetNetworkInterface() isn't clear about what interface is being picked).

bfjelds: Thanks for the help!

The background for this is that to replace Failover Cluster, I'm trying to move a static ip address from one cloud agent to another based on who is the current winner of an election.  To do that, I'm using the PS New-NetIPAddress cmdlet. 

This is method is used when a configuration is not provided, so it is a fallback.  I would be thrilled to have a better (more predictable implementation), but at the time I did not know of one.  Is there a better way to grab the public facing IP address' network interface?

JocelynBerrendonner: Thanks for clarifying! This helps!

So, yes, there is a way to guarantee the interface is always the same. Typically we do that by looking at the route configured on the host (routes are stable over time, unless the admin changes them, which is unlikely). The 0.0.0.0/0 route with the smallest metric is considered the primary public interface. In order to simplify our life, what we typically do in golang is to "Dial" an address on the internet and look at what the local IP address is, then we look for the IP address in the list of all the interfaces. IOW, we let the TCP/IP stack do the route resolution for us. 
The "Dialing" step is implemented in GetIPAddress() in this file. GetIPAddress() will return the local IP of the public facing interface. Once you have it, it should normally be possible to loop through all the network interfaces and find the one that contains that IP (through https://golang.org/pkg/net/#Interface.Addrs).

Hope this helps!

bfjelds: awesome ... thanks!  i tried to do as you suggested and have pushed an update.

rpc/cloudagent/cloud/controlplane/controlplane.proto
```diff
@@ -0,0 +1,40 @@
...
+
+message ControlPlaneRequest {
+	repeated ControlPlane ControlPlanes = 1;
+	Operation OperationType = 2;
+}
+
+message ControlPlaneResponse {
+	repeated ControlPlane ControlPlanes = 1;
+	google.protobuf.BoolValue Result = 2;
+	string  Error = 3;
```
jiria, line 19: nit: extra space

pkg/net/net.go
```diff
@@ -111,3 +112,56 @@ func PrefixesOverlap(cidr1 net.IPNet, cidr2 net.IPNet) bool {
...
+		}
+		// list of unicast interface addresses for specific interface
+		addresses, err := networkInterface.Addrs()
+		if err != nil {
+			return "", err
+		}
+		// network end point address
+		for _, address := range addresses {
+			var ip net.IP
+			switch v := address.(type) {
```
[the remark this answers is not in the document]
bfjelds: i think this is like a type-checked cast?  not sure how to say that in a variable name at this level (where the type isn't known).  do you have a suggestion?

pkg/net/net.go
```diff
@@ -111,3 +112,56 @@ func PrefixesOverlap(cidr1 net.IPNet, cidr2 net.IPNet) bool {
...
+			case *net.IPNet:
```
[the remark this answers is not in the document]
bfjelds: IPNet is the only thing that works on my linux box or my windows machine, but i see a lot of code samples that include both.

rpc/cloudagent/cloud/controlplane/controlplane.proto
bfjelds: 👍

pkg/net/net.go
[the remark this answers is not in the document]
bfjelds: maybe `typedAddress`?

pkg/net/net.go
JocelynBerrendonner: Awesome! Thanks a lot!

JocelynBerrendonner: Nitpicky note: the address returned by GetIPAddress is already guaranteed to be coming from an interface that is up and that is not the loopback. So, the code could be simplified. No biggie though, I am nitpicking.

bfjelds: That makes sense and would definitely simplicy the code.  Would the comparison of IsLoopback would be quicker than comparing 127.0.0.1 to the GetIPAddress result?

PR #14, opened 2020-11-03
Changing Certificate type to string

apluchik:
Changing the certificate types to string from []byte. 
Certificate field in Identity is changed to a map of client addr strings to cert string.


madhanrm (approved):

JocelynBerrendonner (approved):
LGTM

PR #15, opened 2020-11-20
Adding New Way to use the LoginConfig With a hashed CA Cert instead of the Full One

zawachte-msft:
Basically copying the kubeadm join technique.

So today you may see a login config that looks like the following...

```yaml
name: mocuser
token: eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9....7xO6vnKdPahgf1OYXrwnq2f2NNDU-R8UjpQku5Q
certificate: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JS...4rNE9QaFhhUWhFc1pCV0S0tLSk0VSVElGSUNBVEUtLS0tQo=
```

I shortened both the token and the certificate for the example but as you can tell they are both too long to be user-friendly.

For joining a worker node kubadm has a command which looks like this ...

```sh
kubeadm join --discovery-token abcdef.1234567890abcdef --discovery-token-ca-cert-hash sha256:1234..cdef --control-plane 1.2.3.4:6443
```

If they can do something with great UX like this ... then why can't we.

So there are two parts of this to fix.

1. Create a shorter token. This is easy we just need to lower the key length of our JWT tokens.
2. Certificates are a little more tricky. Since our Cloudagent will always be using privately signed certificates (whether we are CA mode or self-signed mode, we will never be going through a public trust process.) cloudagent clients will always need to have the certificate of the cloudagent to be able to trust the cloudagent endpoint is who it says it is. (this is the TLS case, not mTLS.) 

The client needs to check the server certificate is valid. When you go to bing.com you have a trusted third party and roots already installed to your OS, but when you go to cloudagent, unless you locally have its server certificate, how can you trust the endpoint?

This puts a strain on our users because copying around cert files is no fun. This is where kubeadm comes into play. They solved this issue by hashing the server certificate and then taking it as an input. Then from the kubeadm client code they can request the server certificate, hash it, and make sure the hashes match.

So we are doing the same thing here.

This will allow us to get into a world where a user just needs to run

```sh
moctl login --identity --token abcdef.1234567890abcdef  --ca-cert-hash sha256:1234..cdef
```

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -48,6 +48,7 @@ type LoginConfig struct {
 	Name        string
 	Token       string
 	Certificate string
+	CACertHash  string
 }
 
 func (ba *BearerAuthorizer) WithRPCAuthorization() credentials.PerRPCCredentials {
@@ -94,6 +95,8 @@ func NewAuthorizerFromInput(tlsCert tls.Certificate, serverCertificate []byte, s
 	return NewBearerAuthorizer(JwtTokenProvider{}, transportCreds), nil
 }
 
+// NewAuthorizerForAuth is deprecated. Use NewAuthorizerForAuthFromCACertHash as this is the
+// new preferred usage for creating an auth config
 func NewAuthorizerForAuth(tokenString string, certificate string, server string) (Authorizer, error) {
 
 	serverPem, err := marshal.FromBase64(certificate)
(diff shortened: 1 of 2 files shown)


JocelynBerrendonner (approved):

pkg/auth/auth.go
```diff
@@ -114,6 +117,23 @@ func NewAuthorizerForAuth(tokenString string, certificate string, server string)
...
+	pkv := NewPublicKeyVerifier()
+	err := pkv.Allow(cacerthash)
+	if err != nil {
+		return NewBearerAuthorizer(JwtTokenProvider{}, credentials.NewTLS(nil)), fmt.Errorf("could not marshal the server certificate")
+	}
+
+	transportCreds := credentials.NewTLS(&tls.Config{
+		ServerName:            server,
+		InsecureSkipVerify:    true,
+		VerifyPeerCertificate: pkv.VerifyPeerCertificate,
```
apluchik, line 130: Not related to PR; should we look into using VerifyConnection instead?

apluchik (approved):

madhanrm (approved):

pkg/auth/auth.go
zawachte-msft: lets discuss this further.

PR #16, opened 2020-12-11
Adding new certificate fields

apluchik:
Changing certificate value fields which had their types changed back to original types, and instead adding new fields for the new types. This will allow for smoother migration during updates.


madhanrm (approved):

madhanrm:
Good to merge

PR #20, opened 2021-01-07
Dependencies to enable embedded ETCD based config store

jiria:
This is a first change in series that is used to enable embedded etcd based config store. Additional changes are in moc-pkg and wssdcloudagent repos.

This change adds additional properties to the cloudagent entity that are used when adding dynamically additional nodes into the etcd cluster.


rpc/cloudagent/cloud/controlplane/moc_cloudagent_controlplane.proto
```diff
@@ -42,6 +42,14 @@ message ControlPlane {
 	Status status = 6;
 	// state stores the last known election status of this ControlPlane
 	ControlPlaneState state = 7;
+	// etcd client port to listen on
```
madhanrm, line 45: Can you create a separate Object and move these  there

jiria: @madhanrm @bfjelds 

In that case, I think I would create two:

etcdcluster
-id
-name
-locationName
-cacertificate
-cakey
-status

etcdpeer
-id
-name
-etcdClusterName
-locationName
-fqdn
-clientport
-peerport
-status

Let me know if you have any feedback. Also, is there some script for making most of the copying or is it "pick your favority and copy/replace"? Also, thinking about putting these under cloud, but let me know if you have a preference.

jiria: Thinking more, perhaps etcd peer could be just an array property of etcdcluster.

jiria: Decided to keep them split. Let me know if you do not like the approach.

rpc/cloudagent/cloud/etcdcluster/moc_cloudagent_etcdcluster.proto
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache v2.0 license.
```
madhanrm, line 2: If you would like to expose etcd as a service in cloudagent, then the hierarchy has to be built like below.

cloud/etcdcluster
cloud/etcdcluster/peers

Look at Keyvault and other child services under KeyVault

The use case would be users can use cloud agent to create etcd cluster, is what you can get with exposing this 
Like we have kubernetes, etcd would be another service, which can be used.

rpc/cloudagent/cloud/controlplane/moc_cloudagent_controlplane.proto
madhanrm: I was just asking to create a separate Class and not an entire service.

madhanrm: service is also fine, but needs more work

jiria: Aah, misunderstood. I think I have all the pieces ready now, except for the hierarchy. Let me know if you do not see usage for the service and I can fold it back.

jiria: @madhanrm, please let me know if you have a preference whether to keep the service or not.

jiria: Thinking more, if we want to have a two different providers under controlplane for leader election, it does make sense to have these under its own service. So will proceed that way.

rpc/cloudagent/cloud/etcdcluster/moc_cloudagent_etcdcluster.proto
jiria: Just to make sure I understand, did you mean ... and other child services under Group? There do not seem to be child services under KeyVault.

jiria: Ok, took a bit, but I think I have decent idea about the structure of the code and have implemented: etcdserver and etcdcluster are registred under location, and etcdserver is also registered under etcdcluster. One of these days, I would like to learn how did you guys come up with this design. Need to clean it up a bit, will update PR on Monday.

rpc/cloudagent/cloud/etcdcluster/moc_cloudagent_etcdcluster.proto
```diff
@@ -0,0 +1,36 @@
...
+message EtcdClusterResponse {
+	repeated EtcdCluster EtcdClusters = 1;
+	google.protobuf.BoolValue Result = 2;
+	string Error = 3;
+}
+
+message EtcdCluster {
+	string id = 1;
+	string name = 2;
+	string locationName = 3;
```
madhanrm, line 25: We can also keep it under Group, rather than just location, to provide this as service to clients.
This way each clients can get to create their own etcd clusters like k8s clusters

For your purpose, we can keep it Reserved Group under Reserved Location.

rpc/cloudagent/cloud/etcdserver/moc_cloudagent_etcdserver.proto
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the Apache v2.0 license.
+
+syntax = "proto3";
+option go_package = "github.com/microsoft/moc/rpc/cloudagent/cloud";
+package moc.cloudagent.etcdserver;
+
+import "google/protobuf/wrappers.proto";
+import "moc_common_common.proto";
+
+message EtcdServerRequest {
```
madhanrm, line 11: About the hierarchy, I was asking to put etcdserver under etcdcluster, and use etcdcluster as its parent.

The equivalent is Secrets under KeyVault.

So the cli would look like 

mocctl cloud etcdcluster list
modctl cloud ectdcluster server list

jiria: I see, will amend.

rpc/cloudagent/cloud/etcdcluster/moc_cloudagent_etcdcluster.proto
jiria: Ok, will add group into the parent list.

rpc/cloudagent/cloud/etcdserver/moc_cloudagent_etcdserver.proto
jiria: Ok, switched to the model similar to keyvault/secret. I can see that keyvault has secrets, yet I do not see any code updating that. Is that on purpose? @madhanrm

rpc/cloudagent/cloud/etcdcluster/moc_cloudagent_etcdcluster.proto
```diff
@@ -0,0 +1,39 @@
...
+	string groupName = 4;
+	// etcd ca certificate that works as RoT for client and peer connections
+	string caCertificate = 5;
+	// etcd ca key associated with the ca certificate
+	string caKey = 6;
+	repeated EtcdServer servers = 7;
```
madhanrm, line 32: This might not be needed.
If you keep it here, anytime a new server is added, we have to update this in 2 places in cloud agent.

Do you need it or you added it because secrets was under keyvault(not sure why secrets is listed under keyvault - might have to be removed as well)

jiria: Probably do not need it, was mirroring keyvault/secrets. Ok, will simplify.

madhanrm: awesome. Thanks

madhanrm (approved):

PR #22, opened 2021-01-12
Creating IPPools

pradipd:
Adding support for ip ranges in net/net


pkg/net/net.go
```diff
@@ -165,3 +188,76 @@ func GetNetworkInterface() (string, error) {
...
+	if ip.Equal(start) || ip.Equal(end) {
+		return true
+	}
+	if greaterThan(ip, start) && lessThan(ip, end) {
+		return true
+	}
+	return false
+}
+
+//TODO: Create an IPRange class.
```
pradipd, line 263: I can improve all the functions that take a start/end by creating an iprange class.
However, given the time constraints, I'm going to submit like this and when time is available, go back and create an iprange class.

madhanrm (approved):

PR #24, opened 2021-01-27
Certificate helper functions

raghavendra-nataraj:


pkg/certs/certificateAuthority.go
```diff
@@ -0,0 +1,301 @@
...
+	"encoding/asn1"
+	"encoding/pem"
+	"fmt"
+	"math"
+	"math/big"
+	"time"
+)
+
+var (
+	//  1.3.6.1.4.1.311.104  subtree is defined for CIA MSK8S
```
JocelynBerrendonner, line 21: Just out of curiosity, what does "CIA" means in this context? Is it like the name of the agency? :)

pkg/certs/certificateAuthority.go
```diff
@@ -0,0 +1,301 @@
...
+	verifyOptions := x509.VerifyOptions{
+		Intermediates: intermediatesPool,
+		Roots:         ca.rootsPool,
+		KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageAny},
+	}
+
+	_, err = certToRenew.Verify(verifyOptions)
+	if err != nil {
+		if time.Now().After(certToRenew.NotAfter) {
+			return nil, fmt.Errorf("unable to verify certificate to be renewed: Certificate Expired: %v", err)
```
JocelynBerrendonner, line 231: You may want to consider defining an error in our "errors.go" module (https://github.com/microsoft/moc/blob/main/pkg/errors/errors.go), because "certificate expired" is definitely something the calling code can automatically act on (i.e. the calling code may be able to update the certificate).

JocelynBerrendonner: Once you have defined your error, feel free to use errors.Wrapf(<your error>, "unable to verify certificate to be renewed: Certificate Expired: %v", err)

JocelynBerrendonner: Same comment applies to locations below. I think some of these (if not all) may be recoverable by the calling code.

JocelynBerrendonner (approved):
Overall LGTM, but please pay attention to the comments I left. Also, please make sure to submit this once the Jan release is stable.

pkg/certs/certificateAuthority.go
raghavendra-nataraj: I reused code from from cert-tattoo repo. I thought it was (Confidentiality-Integrity-Authentication).

pkg/certs/certificateAuthority.go
raghavendra-nataraj: Sure, It has another revision of the changes. I incorporate the suggested changes into that. Thanks

madhanrm (approved):

pkg/certs/certificateAuthority.go
JocelynBerrendonner: Sounds good! Thanks! :)

pkg/certs/certificateAuthority.go
JocelynBerrendonner: Thanks a lot! Looks good!

PR #25, opened 2021-02-12
Provide sub-folder and config-name options for cloudconfig file

raghavendra-nataraj:


pkg/auth/auth.go
```diff
@@ -279,13 +302,39 @@ func GetWssdConfigLocation() string {
 	}
 	return wssdConfigPath
 }
+
+// GetWssdConfigLocationName gets the path for access filename from environment + subfolder with file name fileName
+func GetWssdConfigLocationName(subfolder, filename string) string {
```
madhanrm, line 307: GetMoc*

Use Moc instead of Wssd

madhanrm (approved):

PR #36, opened 2021-03-09
Protocol buffer definitions for baremetalhost agent

vincepnguyen:


rpc/baremetalhostagent/agent/moc_baremetalhost_agent.proto
```diff
@@ -0,0 +1,114 @@
...
+
+message BareMetalHost {
+	string name = 1;
+	string id = 2;
+	BareMetalHostStorageConfiguration storage = 3;
+	BareMetalHostOperatingSystemConfiguration os = 4;
+	BareMetalHostNetworkConfiguration network = 5;
+	BareMetalHostHardwareConfiguration hardware = 6;
+	Status status = 7;
+	bool disableHighAvailability = 8;
```
cwize1, line 91: High availability isn't a thing for bare-metal machines.

rpc/baremetalhostagent/agent/moc_baremetalhost_agent.proto
```diff
@@ -0,0 +1,114 @@
...
+
+message BareMetalHostHardwareConfiguration {
+	BareMetalMachineSize size = 1;
+}
+
+message BareMetalHostSecurityConfiguration {
+	bool enableTPM = 1;
+}
+
+message BareMetalHost {
```
cwize1, line 83: One of the things I really hate about the ECP code design, is the fact that some of these fields are user provided properties and some of them are reported status values. It would be kind of nice if we could cleanly separate them. Though this would create inconsistency in the ECP codebase, which would also be bad.

rpc/baremetalhostagent/agent/moc_baremetalhost_agent.proto
```diff
@@ -0,0 +1,114 @@
...
+	string name = 1;
```
cwize1, line 84: Do `name` and `id` make sense for the baremetalhost agent, since single agent instance can only handle a single deployment?

rpc/baremetalhostagent/agent/moc_baremetalhost_agent.proto
```diff
@@ -0,0 +1,114 @@
...
+	string customData = 4;
+	repeated BareMetalHostSSHPublicKey publickeys = 5;
+	OperatingSystemType ostype = 6;
+	// bootstrap engine can be cloud-init, Windows answer files, ...
+	OperatingSystemBootstrapEngine osBootstrapEngine = 7;
+	BareMetalHostLinuxConfiguration linuxConfiguration = 8;
+	BareMetalHostWindowsConfiguration windowsConfiguration = 9;
+}
+
+message BareMetalMachineSize {
```
cwize1, line 69: Aren't these values fixed? Why are they settable?

rpc/baremetalhostagent/agent/moc_baremetalhost_agent.proto
jiria: Wonder if a way out of this is to have 2 entities. One tracking the desired state on the machine, the other tracking the machine itself. I recall Madhan mentioned the second entity in one of the earlier conversations as the OsDeployment (my name, but same concept).

rpc/baremetalhostagent/agent/moc_baremetalhost_agent.proto
```diff
@@ -0,0 +1,114 @@
...
+}
+
+message BareMetalHostResponse {
+	BareMetalHost bareMetalHost = 1;
+	google.protobuf.BoolValue Result = 2;
+	string  Error = 3;
+}
+
+service BareMetalHostAgent {
+	rpc CreateOrUpdate(BareMetalHostRequest) returns (BareMetalHostResponse) {}
```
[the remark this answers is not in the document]
jiria: Agent needs to stay, so you can perform runtime operations. An example is A/B update.

rpc/baremetalhostagent/agent/moc_baremetalhost_agent.proto
```diff
@@ -0,0 +1,114 @@
...
```
madhanrm, line 111: The LB is little difference, because it has more entities to control within the lb agent.
In this case, we dont need all these.
Something like https://github.com/microsoft/moc/blob/main/rpc/guestagent/admin/exec/moc_guestagent_exec.proto

For Health and AgentInfo, You can try to reuse or follow 
https://github.com/microsoft/moc/blob/main/rpc/common/admin/health/moc_common_health.proto

Can you list down the use cases so it would be easier to bring in what we want
1. Execute Something
2. Check Notification
3. GetAgentInfo

madhanrm: Instead of execute somehting, you could define APIs can can be specific as well, say Upgrade, Reboot etc as well

rpc/baremetalhostagent/agent/moc_baremetalhost_agent.proto
cwize1: FYI: While bare-metal machines should have two entities in cloud-agent (one for the machine registration and one for the OS deployment), my comment here is a slightly different issue.

This comment is saying that for a given entity in the gRPC API, the settings and the state fields should live in different `message`s. In theory, this will make the API easier to understand.

rpc/baremetalhostagent/agent/moc_baremetalhost_agent.proto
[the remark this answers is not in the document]
cwize1: But once the bare-metal host reboots into the user workload, the baremetalhost agent is no longer running correct?

rpc/baremetalhostagent/agent/moc_baremetalhost_agent.proto
jiria: Sure. Are the messages usually not mirroring the entities though? So if we have separate entities, we would get separate messages (if we follow the usual pattern for messages).

rpc/baremetalhostagent/agent/moc_baremetalhost_agent.proto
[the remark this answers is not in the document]
jiria: We will need to design the flow, but we need to have the agent available on the user workload as well. It can come prebaked in the image or we will need to inject it as part of os deploy. And its state (certs) can live on a separate partition that is not going to get wiped. Actual design of this is still needed. Otherwise we have no good way to perform A/B update.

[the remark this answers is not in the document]
madhanrm: agent will always be running. 
It provides mechanism of 
1. Heartbeat
2. Changes (Notification)
3. Execute something

rpc/baremetalhostagent/agent/moc_baremetalhost_agent.proto
vincepnguyen: I agree. Do you know how I can define the values to be unsettable?

rpc/baremetalhostagent/agent/moc_baremetalhost_agent.proto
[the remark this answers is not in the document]
madhanrm: can we call for a small meeting and close on the gap here

vincepnguyen:
Yes! I set up a meeting for tomorrow to close in on this quickly. Thanks!

rpc/baremetalhostagent/agent/moc_baremetalhostagent.proto
```diff
@@ -14,25 +14,6 @@ message BareMetalHostSSHPublicKey {
...
 }
 
-message BareMetalHostDisk {
-	string diskName = 1;
-	int32 diskSizeGB = 2;
-}
-
-message BareMetalHostStorageConfiguration {
-	repeated BareMetalHostDisk disks = 1;
-	string imageReference = 2;
```
cwize1: I believe this is the OS image that gets deployed. You may need to keep it.

cwize1 (approved):

rpc/baremetalhostagent/agent/moc_baremetalhostagent.proto
```diff
@@ -0,0 +1,82 @@
...
+import "moc_common_common.proto";
+import "moc_common_computecommon.proto";
+
+
+message BareMetalHostSSHPublicKey {
+	string keydata = 1;
+}
+
+message BareMetalHostDisk {
+	string diskName = 1;
```
madhanrm, line 18: What will happen is cloud agent passes in this? Is it going to format and partition the disk?

madhanrm: not sure how this is relevant

rpc/baremetalhostagent/agent/moc_baremetalhostagent.proto
```diff
@@ -0,0 +1,82 @@
...
+	int32 diskSizeGB = 2;
+}
+
+message BareMetalHostStorageConfiguration {
+	repeated BareMetalHostDisk disks = 1;
```
madhanrm, line 23: same as above. What is the functionality that we are exposing with this

rpc/baremetalhostagent/agent/moc_baremetalhostagent.proto
```diff
@@ -0,0 +1,82 @@
...
+	bool enableAutomaticUpdates = 1;
+	string timeZone = 2;
+	BareMetalHostRDPConfiguration RDPConfiguration = 3;
+}
+
+message BareMetalHostOperatingSystemConfiguration {
+	string computerName = 1;
+	BareMetalHostUserConfiguration administrator = 2;
+	repeated BareMetalHostUserConfiguration users = 3;
+	string customData = 4;
```
madhanrm, line 51: I thought the only feature that we need is customData, which has the cloudinit - which could potentially do everything
1. User creation
2. ssh keys 
3. hostname setting
What is the intent is exposing all these properties that cloudagent has to fill and push to this agent

rpc/baremetalhostagent/admin/credentialmonitor/moc_baremetalhostagent_credentialmonitor.proto
```diff
@@ -0,0 +1,30 @@
...
+    string  Error = 2;
+}
+
+message CredentialMonitor {
+    bytes certificate = 1 [deprecated = true];
+    CertificateStatus status = 2; 
+    string newCertificate = 3;
+}
+
+service CredentialMonitorAgent {
```
madhanrm, line 28: Whats the feature/requirement here.

vincepnguyen:
Madhan, how is this credential monitor used for other agents? What is the scenario? Thanks.

rpc/baremetalhostagent/agent/moc_baremetalhostagent.proto
jiria: My thinking was that we will want to turn this the other way around. The bmh agent is reporting to cloud agent the configuration of the bmh so that we could potentially use this information to drive the host selection. We have been talking Vince, Davut, Chris and me about the need to split command&control messages vs reporting messages, but that has not yet happened.

rpc/baremetalhostagent/agent/moc_baremetalhostagent.proto
jiria: Presumably these were mirrors from the VM entity. I believe the logic to generate cloud init was in node agent, not the cloud agent, and so that is probably why these are passed down directly (but you know more here).

rpc/baremetalhostagent/admin/credentialmonitor/moc_baremetalhostagent_credentialmonitor.proto
vincepnguyen: No requirement/feature at the moment. I removed this.

rpc/baremetalhostagent/agent/moc_baremetalhostagent.proto
vincepnguyen: No requirement for now so I removed. We may need them for future AB update feature, we will revisit then.

rpc/baremetalhostagent/agent/moc_baremetalhostagent.proto
vincepnguyen: Remove.

rpc/baremetalhostagent/agent/moc_baremetalhostagent.proto
vincepnguyen: These are mirrors from VM entity to create the provisioning iso drive. I keep the logic to generate cloud-init drive the same as with Compute/VM so I am keeping this struct the same. Depends on the demo condition (whether cloudagent would use only customData, or all), I will revisit this and remove if needed.

vincepnguyen: Remove all of these for now since we only need cloud-init script (customData)

vincepnguyen:
@madhanrm I resolved all your comments. Thanks for the feedbacks. I need to check in this change in to unblock Davut for Cloud agent change. Please feel free to review it and let me know if you have any more feedbacks. I will address them in the next change.

PR #39, opened 2021-03-12
Adding a short error format for Azure consumption

JocelynBerrendonner:
Adding a short error format for Azure consumption.
Azure requires exposing errors as a string in a short format without spaces. Since most of our errors resolve to a non standard string, I am adding a new function to provide this functionality. That said, I am open to feedback!


JocelynBerrendonner:
@madhanrm , if it is OK to break the name of the errors for march, then I can simplify this code. Thoughts?

nagiesek (commented):
LGTM

nagiesek (approved):

PR #44, opened 2021-03-17
Adding certificate expiry check, adding random subject serials for cl…

raghavendra-nataraj:
…ient certificate

### DIFF
--- a/pkg/auth/auth_test.go
+++ b/pkg/auth/auth_test.go
@@ -12,6 +12,9 @@ import (
 	"path/filepath"
 	"testing"
 	"time"
+
+	"github.com/microsoft/moc/pkg/certs"
+	"github.com/microsoft/moc/pkg/errors"
 )
 
 var key *rsa.PrivateKey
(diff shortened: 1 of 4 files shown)


pkg/auth/auth_test.go
```diff
@@ -202,3 +205,59 @@ func Test_GetCertRenewRequiredDelay(t *testing.T) {
...
+		t.Errorf("Failed creating certificate %v", err)
+	}
+
+	x509Cert, err := x509.ParseCertificate(b)
+	if err != nil {
+		t.Errorf("Failed parsing certificate %v", err)
+	}
+
+	certPem := certs.EncodeCertPEM(x509Cert)
+	if err = certCheck(certPem); err == nil || !errors.IsExpired(err) {
```
pradipd, line 260: Should these be 2 separate checks?
Will certCheck() return an error is the cert is expired?
If so, you will never evaluate isExpired() and it will never be tested.

madhanrm (approved):

pkg/auth/auth_test.go
pradipd: Oh, yeah, now I see the certCheck() function :)
you should have 2 separate checks.
1. check that certCheck() returns expire cert.  It can return an error for other reasons
2. check that isExpired returns true.

pkg/auth/auth_test.go
```diff
@@ -202,3 +205,59 @@ func Test_GetCertRenewRequiredDelay(t *testing.T) {
...
+
+	x509Cert, err := x509.ParseCertificate(b)
+	if err != nil {
+		t.Errorf("Failed parsing certificate %v", err)
+	}
+
+	certPem := certs.EncodeCertPEM(x509Cert)
+	if err = certCheck(certPem); err != nil {
+		t.Errorf("certCheck Expected:nil Actual:%v", err)
+	}
```
pradipd, line 234: Do you also want to check that IsExpired() is false?

pkg/auth/auth_test.go
```diff
@@ -202,3 +205,59 @@ func Test_GetCertRenewRequiredDelay(t *testing.T) {
...
+func Test_CertCheckNotExpired(t *testing.T) {
+	now := time.Now().UTC()
+
+	tmpl := x509.Certificate{
+		SerialNumber: new(big.Int).SetInt64(0),
+		Subject: pkix.Name{
+			CommonName: "test",
+		},
+		NotBefore: now,
+		NotAfter:  now.Add(time.Second * 10),
```
pradipd, line 218: Our test machines can be slow.  I'd bump this up much higher.

pradipd (approved):

PR #53, opened 2021-04-09
darwin

raghavendra-nataraj:


pradipd:
Out of curiosity, what is this needed for?

raghavendra-nataraj:
This is for setting acls for files based on OS. os.Chmod does not work in windows as expected

pradipd:
Thanks!

pradipd (approved):

PR #54, opened 2021-04-09
Bare metal host protocol buffer definitions for cloud agent

davutkarabay:
In this PR we split BMMachine into BMMachine and BMHost entities. BMHost entity is the cloud agent provider for the BMHost agent. BMMachine is created on a BMHost. BMHost is location based and BMMachine is group based.


madhanrm:
Do remember to cherry pick this to master. we already forked a branch for our release

madhanrm:
so BMHost is location based and BMMachine is group based?

madhanrm (approved):

davutkarabay:
Yes, BMHost is location based and BMMachine is group based. Updated the description.

rpc/cloudagent/compute/baremetalhost/moc_cloudagent_baremetalhost.proto
```diff
@@ -0,0 +1,85 @@
...
+
+message BareMetalHostHardwareConfiguration {
+	BareMetalHostSize machineSize = 1;
+}
+
+message BareMetalHost {
+	string name = 1;
+	string id = 2;
+	BareMetalHostStorageConfiguration storage = 3;
+	BareMetalHostOperatingSystemConfiguration os = 4;
```
jiria, line 68: Curious, why do we keep OS config on BMH, as opposed to just having it on BMM?

rpc/cloudagent/compute/baremetalhost/moc_cloudagent_baremetalhost.proto
```diff
@@ -0,0 +1,85 @@
...
+	Status status = 10;
+	string locationName = 11;
+	string fqdn = 12; 
+	string certificate = 13;
+	int32  port = 14;
+	int32  authorizerPort = 15;
+	Tags tags = 16;
+}
+
+service BareMetalHostAgent {
```
vincepnguyen, line 83: This is the interface I should use for Bare Metal Host Agent (not the BareMetalMachineAgent interface), right?

vincepnguyen (approved):

rpc/cloudagent/compute/baremetalhost/moc_cloudagent_baremetalhost.proto
davutkarabay: I had a similar thought, but the reason I kept it was primarily for the computerName field which is part of the OS config as we originally copied from Virtual Machine API.

rpc/cloudagent/compute/baremetalhost/moc_cloudagent_baremetalhost.proto
davutkarabay: Correct

rpc/cloudagent/compute/baremetalhost/moc_cloudagent_baremetalhost.proto
jiria: Why do we need to keep computerName on the BMH? Isn't that a property of the OS, hence somewhat transient?

rpc/cloudagent/compute/baremetalhost/moc_cloudagent_baremetalhost.proto
```diff
@@ -0,0 +1,85 @@
...
+message BareMetalHost {
+	string name = 1;
+	string id = 2;
+	BareMetalHostStorageConfiguration storage = 3;
+	BareMetalHostOperatingSystemConfiguration os = 4;
+	BareMetalHostNetworkConfiguration network = 5;
+	BareMetalHostHardwareConfiguration hardware = 6;
+	BareMetalHostPowerState powerState = 7;
+	SecurityConfiguration security = 8;
+	bool claimedByBareMetalMachine = 9;
```
cwize1, line 73: Why not give the name of the bare metal machine, instead of just a true/false value?

rpc/cloudagent/compute/baremetalmachine/moc_cloudagent_baremetalmachine.proto
```diff
@@ -45,34 +33,13 @@ message BareMetalMachineOperatingSystemConfiguration {
...
-}
-
-message BareMetalMachineNetworkConfiguration {
-	repeated BareMetalMachineNetworkInterface interfaces = 1;
-}
-
-message BareMetalMachineSize {
-	int32 cpuCount = 1;
-	int32 gpuCount = 2;
-	int32 memoryMB = 3;
```
cwize1, line 59: [Discussion] It might make sense to have size properties for BMM, and treat it as minimum requirements for the machine. That, cloud agent will choose the smallest BM host that meets the requirements.

It is fine if you want to punt on this for now, as we don't have clear requirements regarding BM host selection.

rpc/cloudagent/compute/baremetalhost/moc_cloudagent_baremetalhost.proto
davutkarabay: I thought a BMH could have a computerName too. Like the computerName when the provisioning OS is running on the bare metal prior to BMM is created.. But I will remove this.

jiria: I dont see a usecase for that, let me know if you do. What would that computername be useful for? I dont think we would use it for IP resolution.

rpc/cloudagent/compute/baremetalmachine/moc_cloudagent_baremetalmachine.proto
davutkarabay: Yes, I will punt this. We had discussed this as a future addition when the requirements are clear. Currently we choose the first BMH found.

rpc/cloudagent/compute/baremetalhost/moc_cloudagent_baremetalhost.proto
davutkarabay: I thought about the same thing, but after implementing the bool everywhere.. which may be useful for debugging as well. I will try to change this.

rpc/cloudagent/compute/baremetalhost/moc_cloudagent_baremetalhost.proto
davutkarabay: Removed it.

jiria (approved):

PR #58, opened 2021-04-27
Add ProviderID helpers, including support for BareMetal.

cwize1:
Add helper functions for handling ProviderID. This ensures that
the ProviderID handling logic is shared for all the components
that interact with MOC ProviderIDs.

In addition, give bare-metal machines their own unique namespace
in the ProviderID ("moc://baremetal/<machine-name>") to ensure
their can't be any name conflicts.

### DIFF
--- a/pkg/providerid/providerid.go
+++ b/pkg/providerid/providerid.go
@@ -1,0 +1,69 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache v2.0 license.
+
+package providerid
+
+import (
+	"fmt"
+	"strings"
+)
+
+const (
+	ProviderName     string = "moc"
+	ProviderIDPrefix string = ProviderName + "://"
+)
+
+type HostType string
+
+const (
+	HostTypeVM        HostType = "vm"
+	HostTypeBareMetal HostType = "baremetal"
+)
+
+func FormatInstanceID(hostType HostType, machineName string) string {
+	switch hostType {
+	case HostTypeVM:
+		// Don't append the host type for VMs to maintain consistency with previous versions.
+		return machineName
+
+	default:
+		return string(hostType) + "/" + machineName
+	}
+}
+
+func FormatProviderID(hostType HostType, machineName string) string {
+	return ProviderIDPrefix + FormatInstanceID(hostType, machineName)
+}
+
+func ParseProviderID(providerID string) (HostType, string, error) {
+	if providerID == "" {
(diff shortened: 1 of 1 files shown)


pkg/providerid/providerid.go
```diff
@@ -0,0 +1,69 @@
...
+
+const (
+	HostTypeVM        HostType = "vm"
+	HostTypeBareMetal HostType = "baremetal"
+)
+
+func FormatInstanceID(hostType HostType, machineName string) string {
+	switch hostType {
+	case HostTypeVM:
+		// Don't append the host type for VMs to maintain consistency with previous versions.
```
romoh, line 26: 👍

pkg/providerid/providerid.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+type HostType string
+
+const (
+	HostTypeVM        HostType = "vm"
+	HostTypeBareMetal HostType = "baremetal"
+)
+
+func FormatInstanceID(hostType HostType, machineName string) string {
```
romoh, line 23: nit - This is the moc machine name and not the real machine name. I'm not sure if there's a way to make that clear through naming or maybe comments.

romoh (approved):

pkg/providerid/providerid.go
```diff
@@ -0,0 +1,69 @@
...
+func FormatProviderID(hostType HostType, machineName string) string {
+	return ProviderIDPrefix + FormatInstanceID(hostType, machineName)
+}
+
+func ParseProviderID(providerID string) (HostType, string, error) {
+	if providerID == "" {
+		return "", "", fmt.Errorf("providerID is empty")
+	}
+
+	withoutPrefix := strings.TrimPrefix(providerID, ProviderIDPrefix)
```
jiria, line 43: Would it make sense to use built in URL parsing logic?

https://golang.org/pkg/net/url/#Parse

madhanrm:
This is moc repo. This has nothing to do with kubernetes.

madhanrm:
> > This is moc repo. This has nothing to do with kubernetes.
> 
> Is there another repo that is shared amongst CAPH, cloud-op, KVA and cloud-provider-moc?

Its fine for now. We can move it later

madhanrm (approved):

pkg/providerid/providerid.go
cwize1: Probably not. While this looks like a URL, it doesn't really follow the URL rules.

PR #74, opened 2021-06-11
Bumping go build version to 1.15.x

apluchik:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/microsoft/moc
 
-go 1.14
+go 1.15
```
madhanrm, line 4: pipeline.yaml changes?

apluchik: The moc/moc-pkg pipelines don't currently have the GoTool task in their build steps. Should I add that as part of this PR?

madhanrm: Yes, lets add it

apluchik: added GoTool task

madhanrm (approved):

PR #78, opened 2021-06-23
Run command support

aeweston-msft:
## Changes introduced by this PR
- Add RunCommandFailed error
- Add VirtualMachineRunCommandRequest and VirtualMachineRunCommandResponse messages
- Add RunCommand rpc for nodeagent and cloudagent VirtualMachineAgent 
- Note: the new rpc is a breaking change, I could split this into two PRs to avoid this, one for the messages and one for the rpc later


rpc/common/moc_common_computecommon.proto
```diff
@@ -83,4 +83,26 @@ enum VirtualMachineOperation {
 	RESET = 2;
 }
 
+enum VirtualMachineRunCommandExecutionState {
+	ExecutionState_FAILED = 0;
+	ExecutionState_SUCCEEDED = 1;
+	ExecutionState_UNKNOWN = 2;
```
JocelynBerrendonner, line 89: I would put "unknown" as 0, because 0 is frequently the default value.

JocelynBerrendonner: So, if a message is coming to a server/client and the execution state field is missing, then it would automatically be mapped to "unknown" instead of "failed"

JocelynBerrendonner (dismissed):
I left a comment, otherwise it looks good to me!

JocelynBerrendonner (approved):
LGTM

madhanrm (approved):

PR #79, opened 2021-07-21
Winrm config

aeweston-msft:
## Changes introduced by this PR
- Add WinRMConfiguration to WindowsConfiguration and add corresponding messages
- Make RunAsUser and RunAsPassword fields of RunCommand request sensitive


rpc/common/moc_common_computecommon.proto
```diff
@@ -58,6 +58,19 @@ enum VirtualMachineSizeType {
 	Unsupported = 99;
 }
 
+enum WinRMProtocolType {
```
JocelynBerrendonner, line 61: I think we shouldn't expose this to the MOC client. This should be completely transparent for the client.

JocelynBerrendonner: Resolved offline! :)

JocelynBerrendonner (approved):
LGTM

PR #80, opened 2021-07-23
MOC changes for downloading image  from sfs

cleonard-git:
Today images are copied on the MOC and the virtual hard disks are created from these images. 
Details of the feature are called out in the below md:
https://github.com/microsoft/wssdagent/pull/514 

These changes are primarily focused on SFS.


madhanrm (approved):

rpc/cloudagent/compute/galleryimage/moc_cloudagent_galleryimage.proto
```diff
@@ -25,6 +25,24 @@ message GalleryImageResponse {
 	string  Error = 3;
 }
 
+message SFSImageProperties {
+	string catalogName = 1;
```
nwoodmsft, line 29: Majority of these fields are not unique to SFS provider. For example, we support catalogName/Audience/Version/ReleaseName through all of the DownloadSDK providers (including http/local/sfs) and the Parts property is also supported by the http provider.

cleonard-git: HI Nick, Thanks for your comment, the primary focus of the first phase is SFS. Http will be taken in phase 2, I will raise a new PR as the HTTP changes get finalized
The "local" here is the existing support to copy from local vhd. I understand the terminologies in downloadSDK local and this can be confusing. During the HTTP phase 2, we can rename this local to something else.

PR #84, opened 2021-08-05
Integrate bare metal support

jiria:
Introduced providerid as per: https://github.com/microsoft/moc/pull/58
Cleans up baremetalhost and baremetalmachine as per: https://github.com/microsoft/moc/pull/54


vincepnguyen (dismissed):

romoh (approved):

pkg/providerid/providerid.go
```diff
@@ -0,0 +1,69 @@
...
+	if len(split) == 1 {
+		// VMs don't have the host type prefix.
+		return HostTypeVM, split[0], nil
+	}
+
+	hostType := HostType(split[0])
+	machineName := split[1]
+
+	if hostType != HostTypeBareMetal {
+		return "", "", fmt.Errorf("providerID contains unknown host type: %s", string(hostType))
```
madhanrm, line 65: General comment: 
try to wrap errors using known errors defined in errors pkg.
You could use NotFound, or define Unknown etc

rpc/cloudagent/compute/baremetalhost/moc_cloudagent_baremetalhost.proto
```diff
@@ -0,0 +1,76 @@
...
+message BareMetalHostNetworkInterface {
+	string networkInterfaceName = 1;
+}
+
+message BareMetalHostNetworkConfiguration {
+	repeated BareMetalHostNetworkInterface interfaces = 1;
+}
+
+message BareMetalHostSize {
+	int32 cpuCount = 1;
```
madhanrm, line 47: can we keep this as uint

rpc/cloudagent/compute/baremetalhost/moc_cloudagent_baremetalhost.proto
```diff
@@ -0,0 +1,76 @@
...
+
+message BareMetalHost {
+	string name = 1;
+	string id = 2;
+	BareMetalHostStorageConfiguration storage = 3;
+	BareMetalHostNetworkConfiguration network = 4;
+	BareMetalHostHardwareConfiguration hardware = 5;
+	BareMetalHostPowerState powerState = 6;
+	SecurityConfiguration security = 7;
+	string bareMetalMachineName = 8;
```
madhanrm, line 64: is this the hostName? can you keep this as MachineName. BareMetal is redundant

madhanrm: or is this reference to the bareMetalMachine - which doesnt seem right - to have this cyclic dependency.

You can maintain these in the internal structures, when machine gets added to host via AddResource

rpc/common/moc_common_common.proto
```diff
@@ -145,6 +145,7 @@ enum ProviderType {
 	CredentialMonitor      = 31;
 	Logging                = 32;
 	Recovery               = 33;
```
madhanrm, line 147: Revendor here is required.

rpc/cloudagent/compute/baremetalhost/moc_cloudagent_baremetalhost.proto
cwize1: This is a reference to `bareMetalMachine`. It is a read-only field that is populated by MOC for convenience and debugging.

madhanrm: There are patterns to do this - look at how group manages the dependencies in groupsInternal structure

Lets follow the same pattern.

jiria: We have the references in both Internal objects: https://github.com/microsoft/wssdcloudagent/blob/04751dc1fc85c66b5668d7148ae6cb5b455824d1/pkg/cloudagent/services/compute/baremetalhost/internal/baremetalhost.go#L16 and https://github.com/microsoft/wssdcloudagent/blob/04751dc1fc85c66b5668d7148ae6cb5b455824d1/pkg/cloudagent/services/compute/baremetalmachine/internal/baremetalmachine.go#L16. 

With that, do you suggest to delete references from both MOC objects (BareMetalHost and BareMetalMachine) and just keep both references in their Internal objects or are you suggesting something else?

madhanrm: reference is required in proto, only if it needs to be input by the client.
For debugging purpose and other internal notification purpose, internal object implementation is good.

Unless required, do not expose properties in proto.

jiria: With that, should any discovered information about BMH be only part of the internal object? Say, CPU count.

madhanrm: https://github.com/microsoft/moc/blob/main/rpc/cloudagent/cloud/node/moc_cloudagent_node.proto#L39
https://github.com/microsoft/moc/blob/main/rpc/cloudagent/storage/container/moc_cloudagent_container.proto#L37

some sample code on how to propagate node info that is discovered

jiria: Re info, I think I want to leave this to another time. I think as we add inspection support, these will need to be modified as well and so would rather do it in one go.

madhanrm (approved):

pkg/providerid/providerid.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/microsoft/moc/pkg/errors"
+)
+
+const (
+	ProviderName     string = "moc"
+	ProviderIDPrefix string = ProviderName + "://"
+)
+
+type HostType string
```
madhanrm, line 17: these enums can be defined in Proto, if required.

madhanrm (approved):

pkg/providerid/providerid.go
jiria: @cwize1 FYI

PR #101, opened 2021-11-09
Fix the issue causing the multinode yamltests to fail

in3xes:
The GetTagValue() function returns the first occurence of a key/value.
Instead it should return the last occurrence

### DIFF
--- a/pkg/tags/tags.go
+++ b/pkg/tags/tags.go
@@ -22,11 +22,17 @@ func AddTag(key, value string, tags *common.Tags) {
 
 //GetTagValue
 func GetTagValue(key string, tags *common.Tags) (string, error) {
+	var ret string
 	for _, tag := range tags.GetTags() {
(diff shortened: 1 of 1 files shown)


pkg/tags/tags.go
```diff
@@ -22,11 +22,17 @@ func AddTag(key, value string, tags *common.Tags) {
 
 //GetTagValue
 func GetTagValue(key string, tags *common.Tags) (string, error) {
+	var ret string
 	for _, tag := range tags.GetTags() {
 		if tag.GetKey() == key {
-			return tag.GetValue(), nil
+			ret = tag.GetValue()
 		}
```
madhanrm, line 29: Not sure if we have duplicate values.

@raghavendra-nataraj - can you take a look

in3xes:
Closing this PR as this is not a correct fix for the pipeline issue

PR #103, opened 2021-11-18
Moc-common, moving vm sizes contants from moc-pkg.

acraghav:


rpc/common/moc_common_computecommon.pb.go
```diff
@@ -245,6 +245,131 @@ var VirtualMachineSizeType_value = map[string]int32{
 	"Unsupported":         99,
 }
 
+type GpuAssignType = int32
```
madhanrm, line 248: This is auto generated code. You cannot edit it.

If you want code to appear here, you should try to update the proto and do make gen

acraghav: This is been moved to moc/common.

madhanrm (approved):

PR #113, opened 2022-03-01
Add new Key field to allow disable key rotation (for ASZ)

flowerbunny1988:
For ASZ worker stamp 22H2 release. The requirement is to not rotate the AES key created in MocStack.
Currently MocStack will rotate key by default. This PR added a new field to allow diabling key rotation. A following PR that contains changes in cloud agent will be sent.


raghavendra-nataraj (approved):

rpc/cloudagent/security/keyvault/key/moc_cloudagent_key.proto
```diff
@@ -48,6 +48,7 @@ message Key {
 	JsonWebKeyCurveName curve = 10;
 	repeated KeyOperation keyOps = 11;
 	Tags tags = 12;
+	bool disableKeyRotation = 13;
```
madhanrm, line 51: instead of disableKeyRotation, can we introduce rotationFrequencyInDays, where 0 would mean, its disabled

flowerbunny1988: There is a global variable AESKeyRotationPeriod in cloudagent that specifies key rotation duration in seconds, hence is using a boolean here to avoid conflict. If we want to go with keyRotationFrequencyInDays, we need to define rules and well document it for people to use it. For example:
1. if keyRotationFrequencyInDays is 0, then disable key rotation
2. if keyRotationFrequencyInDays is negative, then use AESKeyRotationPeriod in cloudagent
3. otherwise use keyRotationFrequencyInDays , and ignore AESKeyRotationPeriod in cloud agent
thought?

flowerbunny1988: @madhanrm any update?

madhanrm: Yes thats what am asking for.
We can bring in keyRotationFrequencyInSeconds and provide the functionality of rotation per key configurable, instead of just disabling it.
Also my suggestion would break existing clients, so 0 cannot be used for disable, where as 0 can be used for GlobalSettings, Negative value can mean disable and Positive value could mean override global.

flowerbunny1988:
Close this PR. New PR sent out:
https://github.com/microsoft/moc-sdk-for-go/pull/98

PR #115, opened 2022-03-08
add keyRotationFrequencyInSeconds to cloudagent key attribute

flowerbunny1988:
This PR added a new field to allow configuration key rotation frequency.

Design doc: https://github.com/microsoft/wssdagent/pull/618

Testing:
New test cases added to WssdTesting and being tested locally that covers:
1. If key rotation frequency is set to -1, then the AES key does not rotate
2. If key rotation frequency is set to positive value, then the AES key rotates

OS tracking task for ASZ: [Task 37702743](https://microsoft.visualstudio.com/OS/_workitems/edit/37702743): ASZ R1: Implement the option to not do key rotation


madhanrm (approved):

madhanrm:
Good to merge

PR #117, opened 2022-03-17
downloadStatus protobuf & pkg/status update 

pathuang-sia:
- Removing errorDetails and state from downloadStatus based on discussion with idc team
- Adding sizeInBytes to galleryImage due to the need of surfacing vhd size from bottom to top (operator level)
- Reverting progress back to int after discussion and renaming to progressPercentage to better reflect the purpose
- Updating pkg/status to remove the removed/updated downloadStatus fields


rpc/cloudagent/compute/galleryimage/moc_cloudagent_galleryimage.proto
```diff
@@ -65,6 +65,7 @@ message GalleryImage {
 	ImageSource sourceType = 13;
 	HyperVGeneration hyperVGeneration = 14;
 	CloudInitDataSource cloudInitDataSource = 15;
+	int64 size = 16;
```
manoj-kadam, line 68: The name should include the unit as well. sizeInBytes?

rpc/common/moc_common_common.proto
```diff
@@ -64,10 +64,8 @@ message ProvisionStatus {
 }
 
 message DownloadStatus {
-	string progress = 1;
+	int64 progress = 1;
```
manoj-kadam, line 67: Rename the progress to progressPercentage as it's not generic progress anymore.

rpc/cloudagent/compute/galleryimage/moc_cloudagent_galleryimage.proto
pathuang-sia: Since node agent vhd has it as size only too, I thought it might be better to align them. Do you still prefer sizeInBytes?

manoj-kadam: Yes

manoj-kadam (approved):

madhanrm (approved):

PR #124, opened 2022-04-28
User/nishinde/keyimportexport

niranjanshnd:
Making changes to the RPC objects to enable key import/export operations. We need to implement these APIs to enable disaster recovery and inter-cluster VM migration. Design documentation can be found at: https://github.com/microsoft/wssdagent/blob/user/nishinde/keyimportexport/doc/cloudagent-key-import-export.md


raghavendra-nataraj (approved):

rpc/common/moc_common_common.proto
```diff
@@ -12,6 +12,8 @@ enum Operation {
 	POST = 1;
 	DELETE = 2;
 	UPDATE = 3;
+	IMPORT = 4;
```
madhanrm, line 15: can we try to reuse Create for Import.

rpc/common/moc_common_common.proto
```diff
@@ -32,6 +34,9 @@ enum ProvisionState {
 	DEPROVISIONED = 14;
 	DEPROVISION_FAILED = 15;
 	DELETE_PENDING = 16;
+	IMPORTING = 17;
+	IMPORTED = 18;
+	IMPORT_FAILED = 19;
 }
```
madhanrm, line 40: Can you also introduce state for export?

rpc/common/moc_common_common.proto
niranjanshnd: IMO, the create and import operations, as they are implemented right now are different enough that they should require separate APIs. The create operation does not allow caller to specify the key whereas the import operation allows caller to specify it. Merging them together will remove the possibility of controlling access to specific operation using RBAC. 
The approach of keeping import and create as separate operations is consistent with what Azure Key Vault does today:
image

madhanrm (approved):
Good to merge.

Please make a release and revendor all other PRs to point to that release.

PR #129, opened 2022-09-01
Added an error type [Attachment Failed]

abhadoriya26:
Here I have added a new error type [AttachmentFailed]. This can be returned whenever any VHD's or ISO files fail to get attached during VM creation.


pkg/errors/errors.go
```diff
@@ -116,6 +117,8 @@ func GetErrorCode(err error) string {
 		return "Delete Failed"
 	} else if IsDeletePending(err) {
 		return "Delete Pending"
+	} else if IsAttachmentFailed(err) {
+		return "Attachment Failed"
```
in3xes, line 121: Could you raise the moc-pkg PR as well? I would like to see how this error is being used. Wondering if we can make this error more generic.

abhadoriya26: Sure

madhanrm (approved):

in3xes:
@abhadoriya26 could you please abandon this PR?

PR #130, opened 2022-09-01
[CensusHost][Moc] - Added additional telemetry to moc

dipalikaluskar123:
As part of Moc telemetry, we needed the following fields in Processor and OperatingSystem protobuf structures. These fields can then be accessed in wmi, moc-pkg, cloudagent and nodeagent code.

New fields added in Processor structure: -
1. Hypervisorpresent    bool        
2. Manufacturer         string       
3. Cputype              uint32     
4. Architecture         uint32

Created new structure for capturing Operating System information: -
 OperatingSystemInfo
1. Operatingsystemsku

Sample Output; This is Single Node Windows Server machine.

image


rpc/common/moc_common_nodeinfo.proto
```diff
@@ -14,7 +14,11 @@ message Processor {
 	string speed = 3;
 	ProcessorType type = 4;
 	bool virtualization = 5;
-	uint32 logicalprocessors = 6;
+	bool hypervisorpresent = 6;
+	uint32 logicalprocessors = 7;
```
JocelynBerrendonner, line 18: You can't change the position of "logicalprocessors" without doing a breaking change. Can "logicalprocessors" be moved back to position 6?

rpc/common/moc_common_nodeinfo.proto
```diff
@@ -27,18 +31,23 @@ message GPU {
...
+}
+
 message NodeInfo {
 	string name = 1;
 	string id = 2;
 	Resources capability = 3;
 	Resources availability = 4;
-	OperatingSystemType ostype = 6;
-	Status status = 7;
+	OperatingSystemType ostype = 5;
```
JocelynBerrendonner, line 43: Same here. Can you keep the previous values to avoid introducing a breaking change?

rpc/common/moc_common_nodeinfo.proto
dipalikaluskar123: Yes, I have reverted it back to 6

rpc/common/moc_common_nodeinfo.proto
dipalikaluskar123: Yes, I updated it to previous values.

randy-danielson (dismissed):

rpc/common/moc_common_nodeinfo.proto
```diff
@@ -27,6 +31,10 @@ message GPU {
 	uint32 count = 3;
 }
 
+message OperatingSystemInfo {
+	uint64 operatingsystemsku = 1;
```
raghavendra-nataraj, line 35: Can we come up with enums for the operatingsystemsku, manufacturer, cputype, architecture?

dipalikaluskar123: We discussed offline (along with Jocelyn, Randy),
For OperatingSystemSku/Architecture, for this PR i.e. September release, we will be just sending in the Sku number and then in Kusto we will do the mapping (i.e. maintain the list in the backend). 
Next PR: We will be adding work-items to create an Xml parser that would generate a list for us and then we can add that list in the Moc code/Kusto.
Similarly, for the other fields.

dipalikaluskar123: Does that sound reasonable ask ? I can surely take it up in the next PR as an action item to add all the enums.

Operating System sku : [WindowsEditions.xml - Repos (visualstudio.com)](https://microsoft.visualstudio.com/OS/_git/os.2020?path=%2Fdata%2FWindowsEditions.xml&version=GBofficial%2Frsmaster&_a=contents)

Architecture here:
https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-processor#properties

raghavendra-nataraj (approved):

rpc/common/moc_common_computecommon.proto
```diff
@@ -129,3 +129,14 @@ message VirtualMachineRunCommandInstanceView {
 	string Output = 3;
 	string Error = 4;
 }
+
+enum Architecture {
```
raghavendra-nataraj, line 133: Please can you add a comment on where you are getting these values from?

dipalikaluskar123: Values fetched from Win32_Processor class.

Documentation here:
https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-processor#properties

raghavendra-nataraj (approved):

rpc/common/moc_common_nodeinfo.proto
```diff
@@ -27,6 +30,10 @@ message GPU {
 	uint32 count = 3;
 }
 
+message OperatingSystemInfo {
```
madhanrm, line 33: Just OperatingSystem is good . No need for Info

rpc/common/moc_common_nodeinfo.proto
madhanrm: if we change the type of this property later, it would be a breaking change.
If we know the list already, I would suggest adding the enum now. The enum values can be expanded later, but closing of type of important.

rpc/common/moc_common_nodeinfo.proto
madhanrm: Lets add OSType as a property

madhanrm: sku can be uint64

madhanrm: https://github.com/microsoft/moc/blob/7bcfbfa2e7ce80a893948f5442734b5b1ba45664/rpc/common/moc_common_computecommon.proto#L29

dipalikaluskar123: OperatingSystemType is already present as part of NodeInfo.
https://github.com/microsoft/moc/blob/7bcfbfa2e7ce80a893948f5442734b5b1ba45664/rpc/common/moc_common_nodeinfo.proto#L42

dipalikaluskar123: > Lets add OSType as a property

Added it..

dipalikaluskar123: > Just OperatingSystem is good . No need for Info

renamed.

madhanrm (approved):

raghavendra-nataraj (approved):

PR #137, opened 2022-11-09
added tags field on VHD

yeonghunki:


in3xes:
@yeonghunki are you planning on adding the tags for storage containers as well?

yeonghunki:
@in3xes Yes, I am. But I'm now sure how can I verify if I added tags to container object properly. Because I don't know if I can update container's tag from moccli like virtual machines and virtual hard disk

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/stretchr/testify v1.5.1
 	golang.org/x/net v0.0.0-20220909164309-bea034e7d591 // indirect
 	golang.org/x/sys v0.0.0-20220915200043-7b5979e65e41 // indirect
+	golang.org/x/text v0.4.0 // indirect
```
yeonghunki, line 15: This was automatically updated from `make generate` regardless of my change

pkg/logging/redirectstderr_linux.go
```diff
@@ -1,7 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the Apache v2.0 license.
 
-//+build unix
+//go:build unix
```
yeonghunki, line 4: this and below `_windows.go` changes were automatically updated from `make all`

go.mod
in3xes: This is okay

pkg/logging/redirectstderr_linux.go
in3xes: This is coming only from 1.17.1?

yeonghunki: I don't think so, I tried same command with go 1.19.2 version and it generated more of this kind of formatting changes

in3xes: typically we include these files. But when I included `redirectstderr_*.go` files, they broke my builds. But let's keep them for now and see if the builds fail for this PR. If they fail, let's fix them.

yeonghunki:
This PR is related with https://github.com/microsoft/moc/pull/139

PR #141, opened 2022-12-09
Reporting exact error in wmi for telemetry

vishal-kadam:
PR as part of this work item - https://dev.azure.com/msazure/msk8s/_sprints/taskboard/ARC-HCI/msk8s/archci/Zinc/2212?workitem=16134858


pkg/errors/errors.go
```diff
@@ -284,6 +287,10 @@ func IsErrDeadlineExceeded(err error) bool {
 	return checkError(err, os.ErrDeadlineExceeded)
 }
 
+func IsWMIError(err error) bool {
+	return checkError(err, WMIError)
```
JocelynBerrendonner, line 291: This won't work. Also, you'll want to move "IsWMIError" to the WMI module (see my comments in the WMI module)

pkg/errors/errors.go
```diff
@@ -47,6 +47,7 @@ var (
 	Unknown              error = errors.New("Unknown Reason")
 	DeleteFailed         error = errors.New("Delete Failed")
 	DeletePending        error = errors.New("Delete Pending")
+	WMIError             error = errors.New("WMI Error")
```
JocelynBerrendonner, line 50: Actually, we don't want to define WMIError here. We want to get the WMI error from the WMI module.

pkg/errors/errors.go
```diff
@@ -47,6 +47,7 @@ var (
...
 )
 
 func GetErrorCode(err error) string {
```
JocelynBerrendonner, line 53: For all the errors defined in the WMI module, we should probably do something like the following in GetErrorCode():

if IsNotFound(err) **|| wmi.IsNotFound(err)** {
    return "NotFound"
}

JocelynBerrendonner: And so on...

pkg/errors/errors.go
```diff
@@ -116,6 +117,8 @@ func GetErrorCode(err error) string {
 		return "Delete Failed"
 	} else if IsDeletePending(err) {
 		return "Delete Pending"
+	} else if IsWMIError(err) {
+		return "WMI Error"
```
JocelynBerrendonner, line 121: You need to return the WMI error text here. So you'll have to do "return err.Error()"

JocelynBerrendonner (dismissed):
LGTM

pkg/errors/errors.go
```diff
@@ -106,16 +107,20 @@ func GetErrorCode(err error) string {
...
 		return "InvalidToken"
-	} else if IsUnknown(err) {
+	} else if IsUnknown(err) || wmi.IsUnknown(err) {
 		return "Unknown"
 	} else if IsDeleteFailed(err) {
 		return "Delete Failed"
 	} else if IsDeletePending(err) {
 		return "Delete Pending"
+	} else if wmi.IsInvalidType(err) {
+		return "Invalid Input"
```
JocelynBerrendonner, line 121: Should be "Invalid Type" here :)
You may still want an additional "else if" for IsInvalidInput(), though

JocelynBerrendonner (approved):
LGTM

PR #143, opened 2023-02-16
Added NodeOutOfCapacityError

HariPrasadV:
Added NodeOutOfCapacityError for cases where a node is out of capacity.

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -28,6 +28,7 @@ var (
 	InvalidVersion       error = errors.New("InvalidVersion")
 	OldVersion           error = errors.New("OldVersion")
 	OutOfCapacity        error = errors.New("OutOfCapacity")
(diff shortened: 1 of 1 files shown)


pkg/errors/errors.go
```diff
@@ -28,6 +28,7 @@ var (
 	InvalidVersion       error = errors.New("InvalidVersion")
 	OldVersion           error = errors.New("OldVersion")
 	OutOfCapacity        error = errors.New("OutOfCapacity")
+	NodeOutOfCapacity    error = errors.New("NodeOutOfCapacity")
```
pradipd, line 31: I think you forgot to push an update.  it should be:
OutOfNodeCapacity

HariPrasadV: Opened a different PR: https://github.com/microsoft/moc/pull/145/files as this was saying that everything is upto date when I tried pushing

PR #144, opened 2023-02-20
added new grpc service

abhadoriya26:


rpc/cloudagent/admin/logging/moc_cloudagent_logging.proto
```diff
@@ -26,3 +28,11 @@ message LogFileResponse {
 	string                      error    = 3;
 	string                      filename = 4;
 }
+
+message SetResponse {
+	bool                        done     = 1;
```
in3xes, line 33: The response should have at least one field for error of string type

rpc/cloudagent/admin/logging/moc_cloudagent_logging.proto
```diff
@@ -26,3 +28,11 @@ message LogFileResponse {
 	string                      error    = 3;
 	string                      filename = 4;
 }
+
+message SetResponse {
+	bool                        done     = 1;	
+}
+
+message SetRequest {
+	uint32 settracelevel = 1;
```
in3xes, line 37: We need to take the location name in the input. We should be able to set the log level for all the node agents.

PR #147, opened 2023-02-22
add inbound nat rule

nijosmsft:
Add inbound nat rules to lb and nic proto buffers. For more details about how its used please see the below cloudagent PR.

[Enabled inbound](https://github.com/microsoft/wssdcloudagent/pull/729)


madhanrm (approved):

rpc/cloudagent/network/loadbalancer/moc_cloudagent_loadbalancer.proto
```diff
@@ -39,6 +46,7 @@ message LoadBalancer {
 	Status status = 10;
 	Tags tags = 11;
 	uint32 replicationCount = 12;
+	repeated LoadbalancerInboundNatRule inboundNatRules = 13;
```
kumarvin123, line 49: Why is LoadbalancerInboundNatRule  embedded in LoadBalancingRule? Both share the same attributes like frontEndPort etc

nijosmsft: it's in LoadBalancer not LoadBalancingRule. looks like github diff is making it a little misleading. If you expand the view it will show up correctly.

PR #151, opened 2023-02-27
added grpc service to set tracelevel

abhadoriya26:


rpc/common/admin/logging/moc_common_logging.proto
```diff
@@ -36,4 +36,15 @@ message LogFileResponse {
 service LogAgent {
 	rpc Get(LogRequest) returns (stream LogFileResponse) {}
 	rpc Rotate(LogRotateRequest) returns (LogRotateResponse) {}
+	rpc Set(SetRequest) returns (stream SetResponse) {};
 }
+
+message SetResponse {
+	bool                        done     = 1;
```
madhanrm, line 43: done is not required

rpc/common/admin/logging/moc_common_logging.proto
```diff
@@ -6,6 +6,14 @@ option go_package = "github.com/microsoft/moc/rpc/common/admin";
 package moc.common.admin;
 import "google/protobuf/wrappers.proto";
 
+enum VerbosityLevels {
```
in3xes, line 9: We are duplicating everything here. Can we keep just one file and use it both cloud and node agent?

rpc/cloudagent/admin/logging/moc_cloudagent_logging.proto
```diff
@@ -26,3 +29,20 @@ message LogFileResponse {
 	string                      error    = 3;
 	string                      filename = 4;
 }
+
+message SetResponse {
```
in3xes, line 33: Can we move these messages to common/admin as well?

PR #155, opened 2023-03-10
1. Bump down go module version 2. remove linking to a private wmi repo

madhanrm:


go.mod
```diff
@@ -1,12 +1,11 @@
 module github.com/microsoft/moc
 
-go 1.19
```
raghavendra-nataraj, line 3: Should we use 1.17 as the version?

apluchik (dismissed):

go.mod
madhanrm: Just trying to keep it lower, so if other modules are linking they dont fail.
I didnt know moc  operator is also using it

raghavendra-nataraj (dismissed):

PR #157, opened 2023-03-17
Add agent uptime information to the GetAgentInfo response

blairurish:
This adds a new property to the GetAgentInfo API allowing clients to query the time in nanoseconds that the cloud agent has been running.


blairurish:
@madhanrm Could you review this? Looks like I don't have permission to add reviewers.
This is what I was thinking for the API we discussed yesterday.

in3xes:
https://github.com/microsoft/moc/blob/main/rpc/common/admin/health/moc_common_health.proto is probably a better place to add these changes.

blairurish:
> https://github.com/microsoft/moc/blob/main/rpc/common/admin/health/moc_common_health.proto is probably a better place to add these changes.

If we did that, would that mean that cloudagent would need to implement the other methods supported by the health endpoint? Looks like it doesn't implement any of them currently. I agree that looks like a better place to put this method though. Also, note that for our use case we really only need this to be on the cloud agent (wasn't sure if this would also require adding the uptime method to the services which do currently support the health endpoint).

in3xes:
> > https://github.com/microsoft/moc/blob/main/rpc/common/admin/health/moc_common_health.proto is probably a better place to add these changes.
> 
> If we did that, would that mean that cloudagent would need to implement the other methods supported by the health endpoint? Looks like it doesn't implement any of them currently. I agree that looks like a better place to put this method though. Also, note that for our use case we really only need this to be on the cloud agent (wasn't sure if this would also require adding the uptime method to the services which do currently support the health endpoint).

my bad, I assumed it's already implemented in the cloud agent. Makes sense to keep it separate. Maybe we can place it here [https://github.com/microsoft/moc/blob/main/rpc/common/admin/health](https://github.com/microsoft/moc/blob/main/rpc/common/admin/health/) just in case we decided to implement `uptime` for the node agents in the future?

rpc/cloudagent/admin/uptime/moc_cloudagent_uptime.proto
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the Apache v2.0 license.
+
+syntax = "proto3";
+option go_package = "github.com/microsoft/moc/rpc/cloudagent/admin";
+package moc.cloudagent.admin;
+
+message UptimeRequest {}
+
+message UptimeResponse {
+	int64 uptime = 1;
```
in3xes, line 11: I think it's a good idea to have an `error` field to handle the failure cases.

in3xes (approved):

rpc/common/admin/uptime/moc_common_uptime.proto
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache v2.0 license.
+
```
madhanrm, line 3: Quick feedback is, I see folks are implementing grpc service for each request.

This is more like AgentInfo and can have more properties. We could later merge version to this.
Can you please rename UptimeRequest to AgentInfo request

blairurish: Note that the HealthAgent service has a method for getting AgentInfo: https://github.com/microsoft/moc/blob/main/rpc/common/admin/health/moc_common_health.proto#L32
Would we want to include this property in the AgentInfoResponse on this service? I think it would make sense, but as I noted in a previous comment, the cloud agent doesn't currently implement this API and at some point we'd need to add support for the uptime property to the other services that implement this API.

madhanrm (approved):
LGTM

PR #164, opened 2023-04-17
Added GRPC services for loglevelling

abhadoriya26:


rpc/cloudagent/admin/logging/moc_cloudagent_logging.proto
```diff
@@ -26,3 +28,19 @@ message LogFileResponse {
 	string                      error    = 3;
 	string                      filename = 4;
 }
```
madhanrm, line 30: curious - why are we making changes in both. Isnt both the agents using the common proto

abhadoriya26: Yes madhan , actually the set service , having setrequest for cloudagent and nodeagent have different parameters

abhadoriya26: So Madhan, Already there were changes for logging in both cloudagent proto and common proto, moc-sdk  was using cloudagant proto and wssd-sdk was using common proto for logging. So that's why I have been consistent with the way which was already there.

madhanrm: okay, lets refactor this later.

madhanrm (approved):

in3xes (approved):

PR #166, opened 2023-04-21
Add Tags to the IpConfiguration Datastructure

gsicilianomsft:
Add Tags value to the IpConfiguration data structure. This will be used to store the NC resource Id.
Also add a function to NetworkInterface to get the primary Ip Configuration from the list of IpConfigurations


nijosmsft (approved):

in3xes (commented):
@madhanrm I'm not sure about adding Tags to `IpConfiguration` here. Could you please take a look?

gsicilianomsft:
The added Tags are being used to store the resourceId/guid assigned to the ipConfiguration within the NetworkController. The guids need to be saved somewhere in Moc to allow update and syncing the ipConfigurations to NC. Guids present in NC that aren't in Moc are deleted for example.

Tags were used here in IpConfiguration because the Tag property was also used in NetworkInterface, and NC uses the same method to update the NetworkInterfaces by storing the NC NIC resourceId in the Tag of Moc's NetworkInterface object.

gsicilianomsft:
@madhanrm, if there are no other comments, this PR will be completed by the middle of this week

nijosmsft:
I am merging this PR. Following are the pipelines executed for this on top of wssdcloudagent.

Moc-cicd (modified with multi-ip tests): [Pipelines - Run 20230821.9 (visualstudio.com)](https://msazure.visualstudio.com/msk8s/_build/results?buildId=78441520&view=results)
Moc-cicd (Master branch): [Pipelines - Run 20230818.5 (visualstudio.com)](https://msazure.visualstudio.com/msk8s/_build/results?buildId=78339002&view=results)
e2e upgrade: [Pipelines - Run 20230816.1 (visualstudio.com)](https://msazure.visualstudio.com/msk8s/_build/results?buildId=78218819&view=results)
AksHci Upgrade CICD Daily: [Pipelines - Run 20230823.5 (visualstudio.com)](https://msazure.visualstudio.com/msk8s/_build/results?buildId=78569319&view=results)
akshci e2e cicd weekly (sdn skipped): [Pipelines - Run 20230823.5 (visualstudio.com)](https://msazure.visualstudio.com/msk8s/_build/results?buildId=78569319&view=results)
Failures are consistent with state of the pipeline
akshci e2e cicd weekly (sdn only): [Pipelines - Run 20230822.2 (visualstudio.com)](https://msazure.visualstudio.com/msk8s/_build/results?buildId=78498267&view=results)
Failures are consistent with state of the pipeline

PR #176, opened 2023-06-02
Adding SecurityType to ProtoBuf

SethBe:
Adding Security Type to protobufs. This is a foundational element used by all components and will be built upon


rpc/cloudagent/compute/virtualmachine/moc_cloudagent_virtualmachine.proto
```diff
@@ -96,9 +96,16 @@ message UefiSettings {
 	bool secureBootEnabled = 1;
 }
 
+enum {
```
SethBe, line 99: bug, accidently ctrl-z will fix

danmamsft (approved):

rpc/cloudagent/compute/virtualmachine/moc_cloudagent_virtualmachine.proto
```diff
@@ -96,9 +96,16 @@ message UefiSettings {
 	bool secureBootEnabled = 1;
 }
 
+enum SecurityType {
```
raghavendra-nataraj, line 99: Can we add this to https://github.com/microsoft/moc/blob/main/rpc/common/moc_common_computecommon.proto. This way we can avoid duplicate definitions

SethBe: Makes sense ill give it a go

SethBe: Note this triggers changes in other repros, i will update those Pr's after this completes

rpc/common/moc_common_computecommon.proto
```diff
@@ -154,3 +154,9 @@ enum Architecture {
 extend google.protobuf.FieldOptions {
 	bool sensitivecompute = 50002;
 }
+
+enum SecurityType {
+        NOTCONFIGURED = 0;
+        TRUSTEDLAUNCH = 1;
+        CONFIDENTIALVM = 2;
+}
```
SethBe, line 162: Nit space

rpc/cloudagent/compute/virtualmachine/moc_cloudagent_virtualmachine.proto
SethBe: Pr's are updated

raghavendra-nataraj (approved):

PR #179, opened 2023-06-12
Moc protobuf change for checking TVM capabilities

danmamsft:
This is the first change for exposing TVM capabilities.
New properties are added to MOC rpc based on this design:
https://github.com/microsoft/wssdagent/blob/master/doc/features/AzureStackHCI/tvm.md

There will be followup PRs to in moc-pkg and node agent after this is checked in
I have coded up in moc-pkg to fill in these properties and tested node agent can get the properties properly.


SethBe (approved):
:shipit:

rpc/common/moc_common_nodeinfo.proto
```diff
@@ -33,6 +33,7 @@ message GPU {
 message OperatingSystem {
 	uint64 operatingsystemsku = 1;
 	OperatingSystemType ostype = 2;
+	string operatingsystembuildnumber = 3;
```
raghavendra-nataraj, line 36: This is looking different from the design doc. Is there a reason for the difference?

danmamsft: The OS version is what we want to check which is the OS build number. I named it build number to be consistent with wmi property name (the wmi property name is called build number)
Since the type is simply string, there is no need to create another type and define the type as string.

raghavendra-nataraj: Makes sense. We usually align closely to azure-sdk. This makes the code extensisble when azure adds new feature.

SethBe (dismissed):
:shipit:

raghavendra-nataraj (approved):

PR #184, opened 2023-06-29
Adding osRegistrationStatus

vishal-kadam:
1. Code changes as part of this user story - https://dev.azure.com/msazure/msk8s/_sprints/taskboard/ARC-HCI/msk8s/archci/Gallium/2306?workitem=16575701.
2. Design document - https://github.com/microsoft/wssdagent/pull/762


ggaurav10 (dismissed):

rpc/common/moc_common_nodeinfo.proto
```diff
@@ -44,13 +54,12 @@ message NodeInfo {
...
 	int64 uptime = 8;
+	OperatingSystem osInfo = 9;
 }
 
-message Resources {   
+message Resources {
 	Processor processor = 1;
 	PhysicalMemory memory = 2;
-	GPU gpu = 3;
-	OperatingSystem osInfo = 4;
```
JocelynBerrendonner, line 53: Was this unexpectedly removed? Would this change cause a backwards compatibility issue?

vishal-kadam: @JocelynBerrendonner osInfo is not consumed by anyone at the moment as confirmed by Madhan. So we decided to move it outside of Resources and add it in NodeInfo.

in3xes: vmCapabilities should retain its number, right?

vishal-kadam: @in3xes I'm not so sure about it. Since we're removing one field, I thought using 4 for vmCapabilities should be fine.

in3xes: https://github.com/microsoft/moc/blob/f5b55407a9b230a9cac53169911c679567ffb6df/rpc/cloudagent/network/virtualnetwork/moc_cloudagent_virtualnetwork.proto#L40

I think this is the right way to mark it reserved

vishal-kadam: Done.

in3xes (approved):

PR #185, opened 2023-06-30
Expose deployment id to MOC client

lpdnju:


rpc/common/admin/deploymentid/moc_common_deploymentid.proto
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache v2.0 license.
+
```
madhanrm, line 3: can we add a property in https://github.com/microsoft/moc/blob/7f6cee209fd07860934f8d088167faa9d0c95931/rpc/common/admin/health/moc_common_health.proto#L24  rather than exposing a new service

madhanrm (approved):

PR #212, opened 2023-10-20
User/dhwanishah/validations for proxy

djshah19:


rpc/cloudagent/compute/virtualmachine/moc_cloudagent_virtualmachine.proto
```diff
@@ -155,7 +155,7 @@ service VirtualMachineAgent {
 	rpc RunCommand(VirtualMachineRunCommandRequest) returns (VirtualMachineRunCommandResponse) {}
 }
 
-message HttpProxyConfiguration {
+message ProxyConfiguration {
```
raghavendra-nataraj, line 158: Can we declare it in common, since it is used by cloudagent and nodeagent

pkg/validations/proxy_validation.go
```diff
@@ -0,0 +1,88 @@
...
+	} else {
+		defer response.Body.Close()
+		fmt.Println("Connected successfully to the proxy server")
+	}
+
+	return nil
+}
+
+func ValidateProxyCertificate(certContent string) error {
+	certContent = strings.Replace(certContent, "-----BEGIN CERTIFICATE-----", "", -1)
```
raghavendra-nataraj, line 61: Why do we need this string replace

pkg/validations/proxy_validation_test.go
```diff
@@ -0,0 +1,97 @@
...
+	// Invalid hostname
+	err = ValidateProxyURL("http://akse2e:akse2e@.ceccloud1.selfhost.corp.microsoft.com:3128", caCertString)
+	expectedResult = "Get \"https://mcr.microsoft.com\": proxyconnect tcp: dial tcp: lookup .ceccloud1.selfhost.corp.microsoft.com: no such host: Invalid Input"
+	if err.Error() != expectedResult {
+		t.Fatalf("Test_ValidateProxyURL test case failed. Expected error %s but got %s", expectedResult, err.Error())
+	}
+
+	// Commenting out below 2 unit tests as the sky proxy server is unreachable from ADO machines and wsl
+
+	// // Unreachable proxy server
```
raghavendra-nataraj, line 40: Can we remove uncommented code

rpc/cloudagent/compute/virtualmachine/moc_cloudagent_virtualmachine.proto
djshah19: done!

pkg/validations/proxy_validation.go
djshah19: we need to only validate whether key is base64 encoded or not. If we pass this string validation fails

pkg/validations/proxy_validation_test.go
djshah19: done!

pkg/validations/proxy_validation.go
raghavendra-nataraj: base64 should not have this at the start and end

raghavendra-nataraj: Where are we getting this from

pkg/proxycert/proxy.crt
```diff
@@ -0,0 +1,26 @@
+-----BEGIN CERTIFICATE-----
```
raghavendra-nataraj, line 1: Please can we generate the cert rather than hardcoding it

djshah19: this is a certificate file not hardcoded string

pkg/validations/proxy_validation.go
```diff
@@ -0,0 +1,78 @@
...
+		},
+	}
+
+	// Create a client
+	client := &http.Client{
+		Transport: transport,
+	}
+
+	// Test the HTTP GET request
+	response, err := client.Get("https://mcr.microsoft.com")
```
raghavendra-nataraj, line 48: Please can we make the URL configurable

pkg/validations/proxy_validation_test.go
```diff
@@ -0,0 +1,76 @@
...
+package validations
+
+import (
+	"testing"
+
+	"github.com/microsoft/moc/pkg/certs"
+)
+
+func Test_ValidateProxyURL(t *testing.T) {
+	caCert, _, err := certs.GenerateClientCertificate("ValidCertificate")
```
raghavendra-nataraj, line 12: I assume there are not valid proxy test cases? Success scenario

pkg/certs/util.go
```diff
@@ -418,3 +418,51 @@ func GenerateCertificateRenewRequestSameKey(cert *tls.Certificate) (retCsr []byt
 	retCsr = EncodeCertRequestPEM(x509CertReq)
 	return
 }
+
+func GenerateExpiredClientCertificate(name string) (*x509.Certificate, *rsa.PrivateKey, error) {
```
raghavendra-nataraj, line 422: This is not a util functon we should be including . This is a test function

pkg/validations/proxy_validation.go
```diff
@@ -0,0 +1,78 @@
...
+	"encoding/pem"
+	"fmt"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/microsoft/moc/pkg/errors"
+)
+
+func ValidateProxyURL(proxyURL string, certContent string) error {
```
raghavendra-nataraj, line 18: When/Where are we going to call this function?

pkg/validations/proxy_validation_test.go
djshah19: no, there are none

pkg/validations/proxy_validation.go
djshah19: in moc sdk and cloud agent for url validation

pkg/validations/proxy_validation.go
```diff
@@ -0,0 +1,78 @@
...
+		return errors.Wrapf(errors.InvalidInput, err.Error())
+	} else {
+		defer response.Body.Close()
+		fmt.Println("Connected successfully to the proxy server")
+	}
+
+	return nil
+}
+
+func ValidateProxyCertificate(certContent string) error {
```
raghavendra-nataraj, line 59: We have a similar function, can we reuse this https://github.com/microsoft/moc/blob/main/pkg/auth/certrenew.go#L74

pkg/validations/proxy_validation.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	// Check if url scheme is http or https
+	if parsedURL.Scheme != "http" && parsedURL.Scheme != "https" {
+		return errors.Wrapf(errors.InvalidInput, "Invalid proxy URL. The URL scheme should be http or https")
+	}
+
+	// Create a certificate pool
+	caCertPool := x509.NewCertPool()
+	caCertPool.AppendCertsFromPEM([]byte(certContent))
```
raghavendra-nataraj, line 32: We should split this function into two, validating the URL and testing connection to the proxy

pkg/validations/proxy_validation.go
djshah19: It's not exported

pkg/validations/proxy_validation.go
djshah19: I can rename the function as ValidateProxyUrlAndTestConection

pkg/validations/proxy_validation.go
djshah19: done!

pkg/certs/util.go
djshah19: I moved this function to cert_utils.go

pkg/validations/proxy_validation.go
raghavendra-nataraj: We can export it?

pkg/validations/proxy_validation.go
raghavendra-nataraj: I thought if the function is split you can call it places where you dont need to tests the connection. So it becomes easy to from other repos

pkg/validations/proxy_validation.go
djshah19: yes

pkg/validations/proxy_validation.go
djshah19: Can you give example in which case you'll want to just validate the proxy url and not test the connection?

raghavendra-nataraj: This is a general coding Principle. A function is meant to perform one operation. Is there a reason we don't want to split this?

raghavendra-nataraj: The use cases would be when the code may want to validate the URL but may not be in the line of sight of Proxy. We are writing helper functions here and they should not tailored to particular use case.

pkg/validations/proxy_validation_test.go
raghavendra-nataraj: Lets call this out in the document and we can resolve this comments. Thanks

raghavendra-nataraj (approved):

pkg/validations/proxy_validation_test.go
djshah19: I added a test case

pkg/validations/proxy_validation_test.go
```diff
@@ -0,0 +1,64 @@
...
+func Test_ValidateProxyURL(t *testing.T) {
+	// Empty proxy url
+	_, err := ValidateProxyURL("")
+	expectedResult := "parse \"\": empty url: Invalid Input"
+	if err.Error() != expectedResult {
+		t.Fatalf("Test_ValidateProxyURL test case failed. Expected error %s but got %s", expectedResult, err.Error())
+	}
+
+	// Invalid proxy url
+	_, err = ValidateProxyURL("//akse2e:akse2e@skyproxy.ceccloud1.selfhost.corp.microsoft.com:3128")
```
raghavendra-nataraj, line 22: I thought we started fact that we dont want to test external servers for unit tests. You can setup local server for the tests

djshah19: This is invalid url test. I dont need any proxy server to test this.

raghavendra-nataraj (approved):

PR #218, opened 2023-11-28
Add enums for pause and save vm operations.

huanyushao:
Add enums for pause and save vm operations.

Include changes of the proto file and generated go file.


madhanrm (approved):

jonah-kohn (approved):
LGTM

PR #223, opened 2023-12-01
Adding back function that was removed when rerunning protobuf

pradipd:


pradipd:
We manually added a function to the generated protobuf files.
This is not a good idea, because:
1.It is a build warnning; https://github.com/microsoft/moc/blob/eb9d06a284b65d38705c570d535922998e463d46/gen.sh#L61
2.It is error prone, because if the protobuf definition changes, you will have to manually copy all the manually added function over.

It is easier to just have another file with the manually added functions.

pradipd:
That file will not get overwritten during make generate

PR #225, opened 2023-12-19
Add support for Network Security Groups

vlappenbusch:


rpc/cloudagent/network/networksecuritygroup/moc_cloudagent_networksecuritygroup.proto
```diff
@@ -0,0 +1,63 @@
...
+import "moc_common_networkcommon.proto";
+
+enum Action {
+	Allow = 0;
+	Deny = 1;
+}
+
+enum Direction {
+	Inbound = 0;
+	Outbound = 1; //TODO: Do we allow Both as an option? and add separate inbound/outbound rules?
```
sbgms, line 19: >//TODO: Do we allow Both as an option? and add separate inbound/outbound rules? [](http://example.com/codeflow?start=15&length=79)

Remove comment. 
Both is not supported.

rpc/cloudagent/network/networksecuritygroup/moc_cloudagent_networksecuritygroup.proto
```diff
@@ -0,0 +1,63 @@
...
+	google.protobuf.BoolValue Result = 2;
+	string  Error = 3;
+}
+
+message NetworkSecurityGroupRule {
+	string name = 1;
+	string description = 2;
+	Action action = 3;
+	Direction direction = 4;
+	string sourceAddressPrefix = 5;
```
msftpgayam, line 38: Why do we have `sourceAddressPrefix` and `sourceAddressPrefixes`? similarly `destinationAddressPrefixes`, `sourcePortRanges` and `destinationPortRanges`?

rpc/common/moc_common_common.proto
```diff
@@ -316,6 +317,9 @@ enum ProviderAccessOperation {
 
 	VirtualNetwork_Create           = 3400;
 	VirtualNetwork_Update           = 3401;
+
+	NetworkSecurityGroup_Create 	= 3500;
```
msftpgayam, line 321: How come we don't have Delete?

rpc/cloudagent/network/networksecuritygroup/moc_cloudagent_networksecuritygroup.proto
vlappenbusch: This was set up to match the API of the azure NSGs that we're trying to match, but we're not using the repeated ones in practice - because sourceAddressPrefix or sourcePortRange can be a comma separated list of address prefixes or port ranges, we just reuse that when needed. The repeated versions can be removed if needed

rpc/common/moc_common_common.proto
vlappenbusch: I was following the operations of other SDN options in this list, I didn't see a delete for other SDN options. I can add Delete if needed for this list, but I didn't see it in many other places

rpc/cloudagent/network/networksecuritygroup/moc_cloudagent_networksecuritygroup.proto
msftpgayam: We need to discuss and decide whether we need the sourceAddressPrefixes or not.

rpc/common/moc_common_common.proto
msftpgayam: @apluchik could you check if this is correct? Don't we need `Delete`?

rpc/common/moc_common_networkcommon.proto
```diff
@@ -18,6 +18,8 @@ enum Protocol {
 	All = 0;
 	Tcp = 1;
 	Udp = 2;
+	Icmpv4 = 3;
+	Icmpv6 = 4;
```
msftpgayam, line 22: Could you verify if these are being used somewhere?

msftpgayam (dismissed):

msftpgayam:
Let's not merge this PR until all the PRs are approved.

sbgms (approved):

sbgms (approved):

msftpgayam (approved):

PR #230, opened 2024-01-31
Add IPOutOfRange error

mchen727:
Add IPOutOfRange error to handle case when provided IP is out of range of vippool


raghavendra-nataraj (dismissed):

pkg/errors/errors.go
```diff
@@ -141,6 +142,8 @@ func GetErrorCode(err error) string {
 		return "QuotaViolation"
 	} else if IsMeasurementUnitError(err) {
 		return "MeasurementUnitError"
+	} else if IPOutOfRange(err) {
```
raghavendra-nataraj, line 145: This needs to be fixed to the new function name

mchen727: fixed

raghavendra-nataraj (approved):

PR #236, opened 2024-02-14
Add support for availability sets in MOC

jocardeMSFT:
Adding Availability Set support in Protobuf definitions


rpc/cloudagent/compute/availabilityset/moc_cloudagent_availabilityset.proto
```diff
@@ -0,0 +1,39 @@
...
+import "moc_common_common.proto";
+
+message AvailabilitySetRequest {
+	repeated AvailabilitySet AvailabilitySets = 1;
+	Operation OperationType = 2;
+}
+
+message AvailabilitySetResponse {
+	repeated AvailabilitySet AvailabilitySets = 1;
+	google.protobuf.BoolValue Result = 2;
```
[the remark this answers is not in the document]
jocardeMSFT: VirtualMachineResponse has the same structure. Is there any reason for the deviation i.e. does VirtualMachine use the result in some way that avset would not?

rpc/cloudagent/compute/availabilityset/moc_cloudagent_availabilityset.proto
```diff
@@ -0,0 +1,39 @@
...
+// https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/compute/armcompute/models.go
+message AvailabilitySet {
+    string name = 1;
+    string id = 2;
+    string locationName = 3;
+    string groupName = 4;
+    Status status = 5;
+    Tags tags = 6;
+	int32 platformFaultDomainCount = 7;
+	repeated CloudSubResource virtualMachines = 8;
```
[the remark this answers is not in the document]
jocardeMSFT: We could piecemeal strings into the object to identify member virtual machines, but we chose to create a generic object that could hold enough identifying information to id any resource, and encapsulate the member VMs in that resource.

This follows the same pattern in the Azure sdk for go at the link above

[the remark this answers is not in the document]
msftpgayam: which link? Could you point me to the exact code?

rpc/cloudagent/compute/availabilityset/moc_cloudagent_availabilityset.proto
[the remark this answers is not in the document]
msftpgayam: My bad. Got confused with the SDK response.

rpc/cloudagent/compute/availabilityset/moc_cloudagent_availabilityset.proto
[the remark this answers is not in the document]
jocardeMSFT: https://github.com/Azure/azure-sdk-for-go/blob/0d4ed404052e8a8238b4d9a56bcd7bb5dbf7b9f4/sdk/resourcemanager/compute/armcompute/models.go#L211C1-L212C32

the link commented into the code above this message

rpc/common/moc_common_common.proto
```diff
@@ -112,6 +112,21 @@ message Status {
 	ValidationStatus validationStatus = 6;
 }
 
+// in the current implementation, cloudagent objects are generally identified by their group name 
+// and name. These fields are used to identify if a POST request should be treated as a create or 
+// an update
+message CloudSubResource {
```
msftpgayam, line 118: Let's move it to `moc_cloudagent_availabilityset.proto` and call it `VirtualMachineConfig` or something similar. I don't think this makes sense to place in common.proto if no other resource is going to use it.

jocardeMSFT: The intent is for a generic message that any subresource can be described in. It is not currently widely in use but it could be to replace all of the ad-hoc config id messages like NetworkInterface and Disk in the VirtualMachine.proto

Just trying to reduce repeated config messages, if you prefer that then I can change it.

msftpgayam: Your point is valid. Again it's the same discussion we had yesterday. Keeping it consistent across all the resources vs bringing in a minor improvement. In this case, I'd say let's keep things consistent so that people are not confused.

rpc/common/moc_common_common.proto
```diff
@@ -112,6 +112,21 @@ message Status {
...
+	string groupName = 1;
+	string name = 2;
+	string realizedName = 3;
```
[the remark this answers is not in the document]
jocardeMSFT: When we update the nodeagent availability set, the virtual machines need to be referenced by realized names so they can be found in the nodeagent store.

This just provides a way to pass down the realized name to the nodeagent avset.

We originally just wanted the cloudagent avset to just reference the name and the nodeagent avset to just reference the realized name with a translation when calling into the wssd sdk to keep the models cleaner.  We reverted to this because the avset provider cannot call back into the vm provider at wssd sdk call time to do the translation. (the vm provider references the avset provider to add members on VM creation.  The avset provider then cannot call back into the vm provider (no circular dependencies))

rpc/cloudagent/compute/availabilityset/moc_cloudagent_availabilityset.proto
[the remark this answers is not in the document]
jocardeMSFT: @msftpgayam please resolve this comment if you are satisfied with the changes

rpc/common/moc_common_common.proto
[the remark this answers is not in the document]
msftpgayam: Does the realized name show up when we list/print the avset cloud agent object using moccli? The realized name is different from the name in the cloud agent for security purposes. avset should show the realized name when we print the object.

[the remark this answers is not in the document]
jocardeMSFT: Do you mean should not print the realized name?

The idea was to filter out the realized name from the moc api view (the realized name just being internally held and utilized) and only returning the name itself.

@liwonMS for response given she is working in cloudagent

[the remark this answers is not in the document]
msftpgayam: Yes

rpc/common/moc_common_common.proto
```diff
@@ -316,6 +317,9 @@ enum ProviderAccessOperation {
 
 	VirtualNetwork_Create           = 3400;
 	VirtualNetwork_Update           = 3401;
+
+	AvailabilitySet_Create		    = 3500;
```
raghavendra-nataraj, line 321: Don't we need availability set update?

rpc/cloudagent/compute/virtualmachine/moc_cloudagent_virtualmachine.proto
```diff
@@ -103,6 +103,12 @@ message SecurityConfiguration {
 	SecurityType securityType = 3;
 }
 
+message AvailabilitySetReference {
+	string groupName = 1;
+	string name = 2;
+	string realizedName = 3;
```
raghavendra-nataraj, line 109: Nit: Why are we using realized name instead of name?

rpc/cloudagent/compute/availabilityset/moc_cloudagent_availabilityset.proto
```diff
@@ -0,0 +1,45 @@
...
+message AvailabilitySetResponse {
+	repeated AvailabilitySet AvailabilitySets = 1;
+	google.protobuf.BoolValue Result = 2;
+	string  Error = 3;
+}
+
+message VirtualMachineReference {
+	string groupName = 1;
+	string name = 2;
+	string realizedName = 3;
```
raghavendra-nataraj, line 25: Same as below

rpc/cloudagent/compute/availabilityset/moc_cloudagent_availabilityset.proto
```diff
@@ -0,0 +1,45 @@
...
+// avset structure is a flattened version of the model in the Azure sdk for go at
+// https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/compute/armcompute/models.go
+message AvailabilitySet {
+    string name = 1;
+    string id = 2;
+    string locationName = 3;
+    string groupName = 4;
+    Status status = 5;
+    Tags tags = 6;
+	int32 platformFaultDomainCount = 7;
```
raghavendra-nataraj, line 37: nit: indentation

rpc/common/moc_common_common.proto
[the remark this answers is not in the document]
jocardeMSFT: @raghavendra-nataraj see above for realized name

rpc/common/moc_common_common.proto
jocardeMSFT: No we do not

rpc/cloudagent/compute/availabilityset/moc_cloudagent_availabilityset.proto
```diff
@@ -0,0 +1,45 @@
...
+	repeated VirtualMachineReference virtualMachines = 8;
```
raghavendra-nataraj, line 38: Having bidirectional representation is easier for validation. We need to make sure there is only one source of truth and the other reference is based of the source of truth. We should not have two brains updating the references. This could cause many race conditions

rpc/cloudagent/compute/virtualmachine/moc_cloudagent_virtualmachine.proto
jocardeMSFT: See previous thread with Pradeep

rpc/cloudagent/compute/availabilityset/moc_cloudagent_availabilityset.proto
raghavendra-nataraj: Not a change request. Just to call out what to watch out for

jocardeMSFT: We don't have 2 brains.  The source of truth for VM membership is the VM provider.

msftpgayam (approved):

raghavendra-nataraj (approved):

PR #239, opened 2024-02-15
add logicalnetwork type.

chengzy1994:


rpc/common/moc_common_common.proto
```diff
@@ -343,3 +343,15 @@ enum CloudInitDataSource {
 	NoCloud = 0;
 	Azure = 1;
 }
+
+enum NetworkType {
```
raghavendra-nataraj, line 347: Shouldn't we already have this enum?

rpc/nodeagent/network/logicalnetwork/moc_nodeagent_logicalnetwork.proto
```diff
@@ -27,8 +28,9 @@ message LogicalNetwork {
 	string id = 2;
 	repeated LogicalNetworkIpam ipams = 3;
 	Status status = 4;
-	Entity entity = 5;
-	Tags tags = 6;
+	NetworkType type = 5;
```
raghavendra-nataraj, line 31: We should never change the number ordering. It should be added at the end.

chengzy1994: okay

rpc/common/moc_common_common.proto
chengzy1994: we have the virtual network type before, I move it from virtual network to common, so that logical network can use it

rpc/nodeagent/network/virtualnetwork/moc_nodeagent_virtualnetwork.proto
```diff
@@ -40,7 +29,7 @@ message VirtualNetwork {
 	string id = 2;
 	repeated Ipam ipams = 3;
 	Dns dns = 4 [(sensitive) = true];
-	VirtualNetworkType type = 5;
```
raghavendra-nataraj, line 43: What are the consequences of changing this? Do we need to make change in nodeagent? We can avoid that

chengzy1994: I don't think it matters, there is no need to change in nodeagent, just adding a new field

chengzy1994: do we want to ignore it?

chengzy1994: the checks passed, can I get an approval to merge it?

raghavendra-nataraj: @chengzy1994 We would be updating this var in wssdagent right? So changing this would require changing wssdagent code?

chengzy1994: I don't think it requires code change in wssdagent, don't see any code related to this order. @msftpgayam could you help to confirm it? correct me if I miss anything.

msftpgayam: I expect code changes in the node agent if we make this change. This field will not be available to the node agent.

What is the need for changing this field?

PR #243, opened 2024-02-26
add logicalnetwork type.

chengzy1994:


msftpgayam (approved):

rpc/nodeagent/network/virtualnetwork/moc_nodeagent_virtualnetwork.proto
```diff
@@ -35,16 +24,17 @@ message VirtualNetworkResponse {
...
 	string name = 1;
 	string id = 2;
 	repeated Ipam ipams = 3;
 	Dns dns = 4 [(sensitive) = true];
-	VirtualNetworkType type = 5;
 	MacPool macPool = 6;
 	Status status = 8;
 	Entity entity = 9;
 	Tags tags = 10;
+	NetworkType type = 11;
```
raghavendra-nataraj, line 37: Why are we changing this? We should not change this

chengzy1994: The original code use VirtualNetworkType, and I tried to add LogicalNetworkType same as it. But it shows the naming conflicts error. So I moved NetworkType definition to common and both virtual network and logical network can use it.

raghavendra-nataraj: I assume we need to handle upgrade scenario for this case. So it is better to keep the same struct

msftpgayam: I missed this point. Thanks for catching it.

msftpgayam: @chengzy1994 let's not modify the `VirtualnetworkType` field in `VirtualNetwork`. 

Earlier the conflict was coming from using the same names in the enum for Lnet as well. This problem can be fixed by renaming the fields in the enum for Lnets.

For example, the enum for Lnet can look something like this:

enum LogicalNetworkType {
    LnetNAT = 0;
    LnetTransparent = 1;
    LnetL2Bridge = 2;
    LnetL2Tunnel = 3;
    LnetICS = 4;
    LnetPrivate = 5;
    LnetOverlay = 6;
    LnetInternal = 7;
    LnetMirrored = 8;
}

This way we don't have to modify the VirtualNetwork object and get around the naming conflict. Sorry should have caught this earlier.

Also, once we are done with this change, can we run the upgrade pipelines with PRs?

chengzy1994: thanks all, fix it now.

PR #251, opened 2024-03-06
Add NSG support for logical networks

vlappenbusch:
Merge after original NSG PR


rpc/common/moc_common_networkcommon.pb.go
```diff
@@ -351,6 +351,53 @@ func (m *Route) GetMetric() uint32 {
 	return 0
 }
 
+type NsgGroup struct {
```
nijosmsft, line 354: >NsgGroup [](http://example.com/codeflow?start=5&length=8)

why cant we use the existing NetworkSecurityGroup ?

vlappenbusch: We could, but I wanted it to be clear that this is a stub. I can change it if you'd prefer.

nijosmsft: yeah, let's use the existing one. it's a well-established pattern in both networkcontroller and in azure.

nijosmsft: thinking abt it, there is another problem. when we introduce back references, ideally, we would want the NSG proto to have a stub LSubnet resource. but such circular dependencies are not supported in protobuf.
one option is, for the read/write ref we will use a real typed stub resource, like in this case. but for readonly ref (like the one in NSG), we will introduce a new ResourceReference type and use that. 

but with that approach we will be restricting ourselves for the future. there are some scenarios like SLB where a backend pool is read/writeable from both nic and lb.

What about we introduce a common ResourceReference type (similar to moc-sdk-for-go) and use that? 

``` go
ResourceReference
     Location string (optional)
     Group string
     Name string
```

mjethwa-msft (dismissed):

rpc/cloudagent/network/networkinterface/moc_cloudagent_networkinterface.proto
```diff
@@ -40,7 +40,7 @@ message IpConfiguration {
 	repeated InboundNatRule inboundNatRules = 8;
 	Tags tags = 9;
 	NetworkType networkType = 10;
-	string networksecuritygroup = 11;
+	ResourceReference networksecuritygroup = 11;
```
mjethwa-msft, line 43: nit: For consistency consider using the variable name in camel case

rpc/common/moc_common_networkcommon.proto
```diff
@@ -51,4 +51,10 @@ message Route {
 	string nextHop = 1;
 	string destinationPrefix = 2;
 	uint32 metric = 3;
+}
+
+message ResourceReference {
+	string Name = 1;
+	string Location = 2;
+	string Group = 3;
```
mjethwa-msft, line 59: Adding a group here so it is generic for all kinds of resources?

vlappenbusch: yes that was the idea! Nithin mentioned we'll also need this for things like slb in the future

rpc/common/moc_common_networkcommon.pb.go
vlappenbusch: implemented!

nijosmsft: thank you doing this!

nijosmsft (approved):

rpc/cloudagent/network/networksecuritygroup/moc_cloudagent_networksecuritygroup.proto
```diff
@@ -49,10 +49,9 @@ message NetworkSecurityGroup {
 	string name = 1;
 	string id = 2;
 	repeated NetworkSecurityGroupRule networksecuritygrouprules = 3;
-	string groupName = 5;
-	string locationName = 6;
-	Status status = 7;
-	Tags tags = 8;
+	string locationName = 4;
```
msftpgayam, line 52: I'm assuming no customers are using this. Please let me know if my assumption is wrong.

If any production clusters are using NSG, we should not reuse numbers here.

rpc/common/moc_common_networkcommon.proto
msftpgayam, line 59: Why do we have `Group` here if NSGs are scoped to location?

vlappenbusch: This is meant to be a very general thing - we'll need to use something like this in the future for backreferences across NC so I included both location and group so that the reference can use whichever it's parented to. Group isn't used for NSG

rpc/cloudagent/network/networksecuritygroup/moc_cloudagent_networksecuritygroup.proto
vlappenbusch: yes, no one is using NSG at the moment!

rpc/common/moc_common_networkcommon.proto
```diff
@@ -51,4 +51,10 @@ message Route {
 	string nextHop = 1;
 	string destinationPrefix = 2;
 	uint32 metric = 3;
+}
+
+message ResourceReference {
```
raghavendra-nataraj, line 56: I would suggest to be more specific here. Lets use NSG resource reference. 

Follow up question ,do we need a resource reference for NSG since I assume they cant be used across location. So only name is required.

nijosmsft: this is derived from what azure uses, for eg; subresource. we want to have a common reference type to use for various resources, not only nsg.

raghavendra-nataraj: Do we have the azure-sdk reference?

nijosmsft: also to the follow up question, moc has the pattern of using just string name, which imho is confusing.  it makes it difficult to fully qualify resources, say for eg; in case of multi-location if that comes in. so imho, its better to use a explicit resourceref. we also have a bunch work coming in slb etc. where the same pattern will be used.

nijosmsft: > azure-sdk

https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/network/armnetwork/models.go

rpc/common/moc_common_networkcommon.proto
msftpgayam: If it's supposed to be a generic thing, let's put it common_common.proto. If it's specifically meant for nsg, let's remove group and name it appropriately.

vlappenbusch: It's generic in the case of networking - we want to reuse this for other references in SDN, like slb references

msftpgayam: I see, in that case it makes sense in network common but let's rename it and fix the fields. Let's remove group.

msftpgayam: Or we can call it `GroupSubResource` and place it in common_common.proto

rpc/common/moc_common_networkcommon.proto
raghavendra-nataraj: This is an example I am looking at. https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/network/armnetwork/models.go#L11181

msftpgayam: We are looking at the sdk models and trying to bring that to the rpc models.

raghavendra-nataraj: I agree with a reference, but I am looking for specific reference and not a generic one

rpc/common/moc_common_networkcommon.proto
nijosmsft: i am not sure if we need to introduce groupsubresource, locationsubresource etc. just having one is good enough. depending on the scope the fields will be populated. this is a pattern in azure too where they use a single id string which may can be the full subscription/resourcegroup/resource etc or a subset based on the context its used.

rpc/common/moc_common_networkcommon.proto
nijosmsft: yeah, we wanted to use this pattern, but then it will cause a cyclic dependency in protobuf which it doesnt support. subnet will have ref to nsg, and nsg will have ref to subnet(when the back ref is introduced, this is now missing).

rpc/common/moc_common_networkcommon.pb.go
nijosmsft: @raghavendra-nataraj fyi

rpc/common/moc_common_networkcommon.proto
msftpgayam: Again, we are trying to influence rpc based on the SDK models. It may make sense in SDK but it's better to keep things specific in protobufs so that we don't have to write if else conditions in the cloud agent code to figure out if Group resource or Location resource.

nijosmsft: we dont need if else, scope of a resource is well defined.

msftpgayam: Right. Since it's well defined, IMO it makes sense to bring that protobufs so that it brings clarity to the code.

We can keep them generic in the SDK.

rpc/common/moc_common_networkcommon.proto
raghavendra-nataraj: We can limit the reference to have name and location . We can ignore the other properties

nijosmsft: > I agree with a reference, but I am looking for specific reference and not a generic one

all resource references are just stubs, whether its generic or not. we have a generic base one(this is like the inheritance pattern, but protobuf doesnt support it). then for specific scenarios if the generic one doesnt suffice introduce specific ones. but i think 90%(in networking) that we are planning to bring just need a generic stub. what i am trying to avoid is bring a lot different stubs for each scenario.

msftpgayam (approved):

PR #255, opened 2024-03-26
gpu grpc change for aks hybrid

wenjlu-lgy:
Document is here: https://microsoft.sharepoint.com/:w:/t/MSk8s/EX7gIW6PInZHngL2j2ZvAi8Bj3QTGk7arC5kSCm7Cv7jLQ?e=KSh5yb


rpc/common/moc_common_computecommon.proto
```diff
@@ -102,14 +102,16 @@ message WinRMConfiguration {
...
+}
 
-	AssignmentType assignment = 1;
+message Gpu {
+	GpuAssignmentType assignment = 1;
 	int32 partitionSizeGB = 2;
+	int32 gpuCount = 3;
 }
 
 message VirtualMachineCustomSize {
```
dhavalpopat, line 117: Can you remove gpuList from VirtualMachineCustomSize now that we are adding it in the HardwareConfiguration?

wenjlu-lgy: Removed

rpc/common/moc_common_nodeinfo.proto
```diff
@@ -61,7 +62,7 @@ message Resources {
 	reserved 4;
 	Processor processor = 1;
 	PhysicalMemory memory = 2;
-	GPU gpu = 3;	
+	GPU gpu = 3;	// delete it after gpuList is applied
 	VirtualMachineCapabilities vmCapabilities = 5;
-	repeated Gpu gpuList = 6;
+	repeated GPU gpuList = 6;
```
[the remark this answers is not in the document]
wenjlu-lgy: I changed the GPU in node info rather than the one in compute common. Actually they have similar contents, it can be changed to  the one in compute common.

dhavalpopat (approved):

rpc/common/moc_common_nodeinfo.proto
```diff
@@ -33,10 +33,12 @@ message PhysicalMemory {
 	uint64 sizeBytes = 1;
 }
 
-message GPU {
+message HostGPU {
 	string name = 1;
 	uint32 sizeBytes = 2;
-	uint32 count = 3;
+	uint64 partitionSizeBytes = 3;
```
raghavendra-nataraj, line 39: This would break the protobuf contract?
Any removal of params should not reuse the field id. We need to call it out as deprecated and use new field

rpc/common/moc_common_computecommon.proto
```diff
@@ -102,21 +102,23 @@ message WinRMConfiguration {
 	repeated WinRMListener Listeners = 1;
 }
 
-message Gpu {
-	enum AssignmentType {
-		GpuDDA = 0;
-		GpuP = 1;
-	}
+enum GpuAssignmentType {
```
raghavendra-nataraj, line 105: What impact does this have on the protobuf contract?

wenjlu-lgy: For hybridaks, cloud operator will call cloud agent to get the cluster info (especially GPU Assign Type) to decide the plugins to install.

dhavalpopat: @raghavendra-nataraj If you were asking about the change from AssignmentType to GpuAssignmentType, the AssignmentType enum is not being used anywhere, it was added by my earlier PR. So I don't think there should be any issue with protobuf contract

raghavendra-nataraj: When you say earlier. A release is been made with this protobuf. So I believe it might be impacted on the structure

dhavalpopat: Oh got it, although the GPU struct is not being used in the new release, so it shouldn't matter, right? I spoke with Pradeep during office hours, and he believes it shouldn't impact either.

Pradeep brought a point that it may affect the marshaling of VM, but it is not very likely. I can test and verify this needed.

Regardless, if this feels like it could affect the existing structure, then we could keep using the same name for AssignmentType enum.

raghavendra-nataraj: Yes , lets test this scenario. We hit a similar issue in SDN where the we assumed that it as long as the field is not used. but we hit issues during marshalling

dhavalpopat: I tested with the older release, and it seems there is some issue after replacing the binaries from this moc PR. We will revert the name change.

rpc/common/moc_common_nodeinfo.proto
```diff
@@ -58,10 +61,11 @@ message NodeInfo {
...
 message Resources {
+	reserved 3;
 	reserved 4;
+	reserved 6;
 	Processor processor = 1;
 	PhysicalMemory memory = 2;
-	GPU gpu = 3;	
 	VirtualMachineCapabilities vmCapabilities = 5;
-	repeated Gpu gpuList = 6;
+	repeated HostGPU gpuList = 7;
```
raghavendra-nataraj, line 70: I dont think we need to realign this to 7. Since it is just the message name changing

dhavalpopat: It's actually a field change. message name has been changed from "GPU" (uppercase) to "HostGPU". But field number 6 was "Gpu"

Sorry for the confusion. I changed the upper-case GPU to HostGPU so that we can avoid the name confusion in the future.

msftpgayam (approved):

raghavendra-nataraj (approved):

common/virtualmachineconstants.go
```diff
@@ -32,7 +32,7 @@ type VmSize struct {
 	GpuName       string
 	MemoryMB      int
 	GpuAssignMode GpuAssignType
-	GpuList       []*pb.Gpu
+	GpuList       []*pb.VirtualMachineGPU
```
madhanrm, line 35: Is this a breaking change?

dhavalpopat: It's the `.go` file so we don't have to worry about marshaling/unmarshaling issues. Although since this gets used in `moc-pkg`, it will definitely break the current code. I have a `moc-pkg` PR to use this new object.

https://github.com/microsoft/moc-pkg/pull/502

PR #258, opened 2024-04-05
Fix NSG deserialization

vlappenbusch:


rpc/cloudagent/network/networkinterface/moc_cloudagent_networkinterface.proto
```diff
@@ -40,7 +41,7 @@ message IpConfiguration {
 	repeated InboundNatRule inboundNatRules = 8;
 	Tags tags = 9;
 	NetworkType networkType = 10;
-	NetworkSecurityGroupReference networksecuritygroup = 11;
+	NetworkSecurityGroupReference networkSecurityGroupRef = 12;
```
mjethwa-msft, line 44: Can you briefly explain why do we need to do this?

msftpgayam (approved):

rpc/cloudagent/network/networkinterface/moc_cloudagent_networkinterface.proto
vlappenbusch: We tried just renaming, but it seems like something in deserialization will still pick up that they're the "same thing". The issue that we're resolving is I mistakenly thought we hadn't released with NSG without lnet but we had. because of that, if someone had saved data from a March build and tried to upgrade, their data wouldn't deserialize, for NSG, VNIC or VNET. By changing the name, and the message number (and keeping the existing message number reserved), we can make sure anyone who upgrades will keep their data.

mjethwa-msft: got it! thanks!

mjethwa-msft (approved):

rpc/cloudagent/network/networksecuritygroup/moc_cloudagent_networksecuritygroup.proto
```diff
@@ -46,12 +46,13 @@ message NetworkSecurityGroupRule {
 }
 
 message NetworkSecurityGroup {
+	reserved 5; // deprecated group name (now parented to location)
```
raghavendra-nataraj, line 49: I am not sure if we need this. Since we did not create NSG in previous version

vlappenbusch: I don't necessarily know if we did, but considering it was released it's possible that someone could have created, so I think the stakes of keeping it reserved are low. If you want to remove the reservation we can, but I'd want to err on the side of being cautious

raghavendra-nataraj (approved):

PR #266, opened 2024-06-14
Precheck Capacity before Provisioning Virtual Machines and Virtual Hard Disks

JamesLiang29:


rpc/cloudagent/compute/virtualmachine/moc_cloudagent_virtualmachine.proto
```diff
@@ -156,7 +156,18 @@ message VirtualMachineRunCommandResponse {
...
 }
 
+message VirtualMachineCapacityCheckRequest {
+	repeated VirtualMachine VirtualMachines = 1;
+}
+
+message VirtualMachineCapacityCheckResponse {
+	google.protobuf.BoolValue Result = 1;
+	string Error = 2;
+	map<string, string> Placement = 3;
```
raghavendra-nataraj, line 166: What does this parameter hold?

rpc/cloudagent/compute/virtualmachine/moc_cloudagent_virtualmachine.proto
```diff
@@ -156,7 +156,18 @@ message VirtualMachineRunCommandResponse {
...
+}
+
 service VirtualMachineAgent {
+	rpc CheckCapacity(VirtualMachineCapacityCheckRequest) returns (VirtualMachineCapacityCheckResponse) {}
```
raghavendra-nataraj, line 170: Should we name this more generic like precheck. I think we can reuse the same to validate other checks like if the gallery image is healthy, storage container is healthy. We can explore on making the status more specific. Thoughts?

JamesLiang29: OK. Will update.

rpc/cloudagent/compute/virtualmachine/moc_cloudagent_virtualmachine.proto
JamesLiang29: Adding comments.

JamesLiang29: Add comments.

raghavendra-nataraj: Thanks @JamesLiang29 . This might not be the same when the VM is created. Is this specific to GPU? 
Also can we add result in terms of what the requested resource is and what the available resources are? Like CPU request and what is available. This would give a clear idea on what failed. Thoughts?

madhanrm: Not sure what is the use case for MOC to Giving out node names and clients taking dependency on it. Can you explain a scenario on how to use this data? We usually do not give this info back to client on the node placement.

Can you give a sample output please?

For instance, if a request comes in and there are several errors on each node, what would be the error looking like and what would the placement look like

madhanrm: Is the error going to be a cumulative error from other entities that VM interacts with?

madhanrm: Are you also looking to provide a summary of the errors, rather than client retrying multiple times.

raghavendra-nataraj (approved):

PR #272, opened 2024-07-12
VirtualHardDisk Download and GalleryImage Create From OS Disk

damanMSFT:


rpc/cloudagent/compute/galleryimage/moc_cloudagent_galleryimage.proto
```diff
@@ -77,6 +77,7 @@ message GalleryImage {
 	ImageSource sourceType = 13;
 	HyperVGeneration hyperVGeneration = 14;
 	CloudInitDataSource cloudInitDataSource = 15;
+	string sourceVM = 16;
```
msftpgayam, line 80: Do we really need this field? We would be storing the path of OS disk in `sourcePath` and ImageSource in `sourceType`. Does this field add any value after the VM is deleted?

rpc/common/moc_common_common.proto
```diff
@@ -331,10 +331,11 @@ extend google.protobuf.FieldOptions {
...
 enum ImageSource {
-        LOCAL_SOURCE = 0;
-        SFS_SOURCE   = 1;
-        HTTP_SOURCE  = 2;
-        CLONE_SOURCE = 3;
+        LOCAL_SOURCE 	= 0;
+        SFS_SOURCE   	= 1;
+        HTTP_SOURCE  	= 2;
+        CLONE_SOURCE 	= 3;
+		VMOSDISK_SOURCE = 4;
```
msftpgayam, line 338: nit: Fix indentation

PR #275, opened 2024-07-31
[Availability Zone] VM Placement in a Zone integration

bgriddaluru:
Design Doc: https://github.com/microsoft/wssdagent/pull/994 (VM_Placement_in_a_Zone.md)

Changes:
1. Add proto buf related changes for VM association to Availability Zone
2. Zone message used in VM proto on cloud agent is kept in common, so that it can be reused by Placement group.
3. Placement Status is added to generic status. Since it is not applicable to all entities hiding it in output if field is not set.


rpc/common/moc_common_common.proto
```diff
@@ -103,13 +103,25 @@ message Version {
 	string number = 1;
 }
 
+enum PlacementStatusType {
+	UNKNOWN_PLACEMENT = 0;
+	ALIGNED = 1;
+	NOT_ALIGNED = 2;
+}
+
+message PlacementStatus  {
```
msftpgayam, line 112: Can we move the placement status to `ZoneConfiguration` ?

rpc/nodeagent/compute/virtualmachine/moc_nodeagent_virtualmachine.proto
```diff
@@ -108,6 +108,16 @@ message SecurityConfiguration {
 	SecurityType securityType = 3;
 }
 
+message ZoneReference {
+	string name = 1;
+	repeated string nodes = 2;
+}
+
+message ZoneConfiguration {
```
msftpgayam, line 116: Looks like there are duplicates in `ZoneConfiguration` declarations with minor differences. Is it possible to merge them?

bgriddaluru: From CLI to cloud agent only zone names are passed. 
Since zones to node mapping  is only on cloud agent side,  when sending Create VM request to node agent , I am including node names associated with zone. 
Hence the minor difference in spec.

Merging them will lead to an unused field (nodes) on cloud agent side. Since it will be reused by placement group spec, Piyush suggested to keep them separate to keep it clean. 

Please let me know if merging is preferred.

rpc/common/moc_common_common.proto
bgriddaluru: Forgot to add couple of reasons for adding it in common place.
1.  Simplify status parsing for above layers 
2. VM placement status is applicable to placement group also. 

So, it will be better to keep it in common status or define a separate placement status field in VM spec. 
Please share your thoughts on this.

rpc/nodeagent/compute/virtualmachine/moc_nodeagent_virtualmachine.proto
msftpgayam: We can keep them separate in this case. Can we add this explanation in the comments so that others are not confused in the future?

rpc/common/moc_common_common.proto
msftpgayam: okay, in that case we can define the placement status in common and reuse it wherever needed

rpc/nodeagent/compute/virtualmachine/moc_nodeagent_virtualmachine.proto
bgriddaluru: Add comments for defining separately for cloud agent and node agent.

Zone configuration on cloud agent will be used by VM and Placement Group, so moved it to common based on feedback from Piyush.

rpc/common/moc_common_common.proto
bgriddaluru: Make it a field for VM or hide the status field for other entities

bgriddaluru: Made changes to hide placement status field if it is not set.

bgriddaluru: Created two VMs one with zone associated and another without. Captured Get-MocVirtualMachine output for both. |
PlacementStatus field is seen only for VM with zone association. Please refer to image below

image

msftpgayam (approved):

piyushka17 (approved):

PR #276, opened 2024-08-05
Add SecurityEncryptionType property for cloudagent gRPC

bigjeffxia2022:
1. Define the protocol between moc operator and cloud agent, so moc operator can proceed to consume the change.
2. The handle of this property will be done by moc team in the following.

Bug: #28372823


JamesLiang29 (approved):

rpc/cloudagent/compute/virtualmachine/moc_cloudagent_virtualmachine.proto
```diff
@@ -33,6 +33,8 @@ message SSHPublicKey {
 message Disk {
 	// reference to the virtual hard disk managed by storage
 	string diskname = 1;
+	// security encryption type for OsDisk. Possible values include: "NonPersistedTPM"
+	string securityEncryptionType = 2;
```
danmamsft, line 37: securityEncryptionType doesn't belong to OSDisk directly, here is the structure in ARM API:
https://github.com/Azure/azure-rest-api-specs-pr/blob/RPSaaSMaster/specification/azurestackhci/AzureStackHCI.StackHCIVM.Management/models.tsp#L1483
We should define in as the same structure in Moc to align with Azure. Here is a signed-off design (including Moc API change): https://github.com/microsoft/wssdagent/pull/941/files

bigjeffxia2022: Thanks, Dan. I will drop this PR.

PR #281, opened 2024-08-16
Add SecurityEncryptionType

shayan-msft:
Adding ManagedDisk, SecurityProfile, and SecurityEncryptionType


msftpgayam (dismissed):

msftpgayam (approved):

rpc/common/moc_common_computecommon.proto
```diff
@@ -237,6 +237,19 @@ enum SecurityType {
         CONFIDENTIALVM = 2;
 }
 
+enum SecurityEncryptionTypes {
+    SecurityEncryptionNone = 0;
+    NonPersistedTPM = 1;
```
SethBe, line 242: Note for other platforms. Stateless is NonManagementVTL. not NonPersistedTPM.
At the VM creation time you will need to determine which type to set based off of the platform.  Just makeing sure that this doesnt limit us.

SethBe (approved):

rpc/common/moc_common_computecommon.proto
shayan-msft: discussed offline, we will revisit renaming this in the future

PR #282, opened 2024-08-19
Adding enhancements for Zones in moc

mkatturu:


go.sum
```diff
@@ -1,945 +1,21 @@
-cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
msftpgayam, line 1: Any idea why a small change to the health state enum caused so many changes here?

rpc/common/moc_common_common.proto
```diff
@@ -56,6 +56,7 @@ enum HealthState {
 	DEGRADED = 5;
 	// The entity went missing from the agent
 	NOTFOUND = 6;
+	NOTREADY = 7;
```
msftpgayam, line 59: Is there a doc to describe what each health state means for zones?

mkatturu: Yes this is defined in our zones feature design doc. 

https://github.com/microsoft/wssdagent/blob/master/doc/features/zones/Zone_Management_in_moc.md

image

piyushka17 (dismissed):

go.sum
mkatturu: I updated my branch and go sum is now updated and almost all the changes in go sum are reverted

jocardeMSFT:
General Comment, request you add more to the PR description including change summary, change purpose, and testing complete. From a maintainability point of view, it helps future devs understand the intent of your changes and adds context for why things were implemented a certain way.

jocardeMSFT (approved):

go.mod
```diff
@@ -20,7 +20,7 @@ require (
 	golang.org/x/net v0.29.0 // indirect
```
piyushka17, line 20: I think you shouldn't have any change in go.mod, go.sum. You can possibly revert these files?

piyushka17 (approved):

PR #288, opened 2024-08-27
VirtualHardDisk Download

damanMSFT:


rpc/common/moc_common_common.proto
```diff
@@ -348,8 +349,9 @@ enum HyperVGeneration {
 } 
 
 enum DiskFileFormat {
-	DiskFileFormatVHDX =  0;
-	DiskFileFormatVHD = 1;
+	DiskFileFormatUNKNOWN = 0;
+	DiskFileFormatVHDX =  1;
+	DiskFileFormatVHD = 2;
```
msftpgayam, line 354: we should only append to the enum. If we change this order and an old object has DiskFileFormat based on the previous values, their meaning will change after the upgrade.

damanMSFT: Replied about this on teams

damanMSFT: Done

msftpgayam (dismissed):

raghavendra-nataraj (approved):

msftpgayam (approved):

rpc/common/moc_common_common.proto
```diff
@@ -350,6 +350,7 @@ enum HyperVGeneration {
 enum DiskFileFormat {
 	DiskFileFormatVHDX =  0;
 	DiskFileFormatVHD = 1;
+	DiskFileFormatUNKNOWN = 2;
```
liwonMS, line 353: Are we expecting cx to actually input Unkonwn. I couldn't really think of situation where this is needed. especially when default value for int32 is 0 (VHDX).

liwonMS: if we dont have cx flow that would allow this value to be passed down, This should be removed.

liwonMS (approved):

PR #289, opened 2024-08-28
SetHealth doesn't overwrite previousHealth unnecessarily

nikohobart:
Avoids scenarios where we get previousState == currentState, which is
1. not helpful
2. often erases valuable information (as we sometimes have multiple calls to this function in a single moc operation)


pkg/status/status.go
```diff
@@ -34,7 +34,9 @@ func SetError(s *common.Status, err error) {
 
 // SetHealth
 func SetHealth(s *common.Status, hState common.HealthState, err ...error) {
-	s.Health.PreviousState = s.Health.CurrentState
+	if s.Health.CurrentState != hState {
```
ganeshamohana, line 37: According to this condition we don't want to change the previous state unless the current state changes to a new state?

What about the following scenario?

1. Current state: Pending
2. Next state: active
3. Perform some operation on the entity
4. Next state: active

While debugging the above operation when we look at the statuses, we will have false information on the previous state for the operation we're trying to debug.

nikohobart: Ah yeah this is the case for ProvisionState, but not HealthState. HealthState has no pending states, so currently what happens is we just get lots of prevState=OK, currState=OK

ganeshamohana: It was an example scenario.

prevState=OK, currState=OK is valid information to know that the health state was good and it's still good after some operations. Consider positive scenarios where the state gets updated tens if not hundred times to "OK". Do we keep the previous state not updated in this scenario? If yes then what's the use of keeping that state info?

ganeshamohana: @msftpgayam

nikohobart: Hm I guess what's the original value of the prevState field? 

Because I'm not sure I see the value in keeping it in the original scenario either. E.g., a vm client create constitutes multiple health status set calls. One each for internal.Create, internal.Start, and internal.Get. If we were concerned about tracking each of the outcomes here to paint a picture of how the calls degraded, then prevState & currState are a poor proxy for that, as they only track the last two calls out of any number of internal service calls--resulting in key info missing. 

The original scenario / why I made the change was in the case of platform monitoring:
Cloudagent receives notification that a container path was added back and updates health state -> a validation function is run on the container object which also updates the health state. Now we just have prevstate==currState==OK. We have no visibility into what occurred before. Vs., with this change, we would know that the container was previously missing, then some change occurred that got it back into an OK state. It is true though, that we wouldn't have a frame of reference on when that shift occurred off of a quick glance.

ganeshamohana: "It is true though, that we wouldn't have a frame of reference on when that shift occurred off of a quick glance."

This is my question, how is this new logic going to help us debug better than the old implementation?

PR #290, opened 2024-09-03
Make HealthState show enum 0 value

nikohobart:


pkg/status/status.go
```diff
@@ -138,3 +145,39 @@ func GetFromStatuses(statuses map[string]*string) (status *common.Status) {
 
 	return
 }
+
+// HealthState requires custom parsing as it is a proto enum. Otherwise enum 0 (NOT_KNOWN) will be an empty string.
+func parseHealth(hstate *common.Health) string {
+	if hstate == nil {
+		return ""
```
msftpgayam, line 152: Can we return health state as "NOT_KNOWN" if the input hstate is nil?

pkg/status/status_test.go
```diff
@@ -0,0 +1,64 @@
+package status
```
msftpgayam, line 1: let's add an unit test for the case where the health state is nil and make sure the reverse conversion is returning NOT_KNOWN.

pkg/status/status.go
```diff
@@ -138,3 +143,39 @@ func GetFromStatuses(statuses map[string]*string) (status *common.Status) {
 
 	return
 }
+
+// HealthState requires custom parsing as it is a proto enum. Otherwise enum 0 (NOT_KNOWN) will be an empty string.
+func parseHealth(hstate *common.Health) string {
+	if hstate == nil {
+		return fmt.Sprintf("currentState:%s", common.HealthState_name[0])
```
msftpgayam, line 150: Can we use the const `HealthState_NOTKNOWN` instead of 0.

nikohobart: oh good point yes done

msftpgayam (approved):

chengzy1994 (approved):

PR #292, opened 2024-09-04
[SDN-VNET]  Protobuf changes

nijosmsft:
Main changes
- Added SubnetReference to IpConfiguration to support SDN VNET 
- Added ApiVersion to VirtualNetworkRequest
- Added LogicalNetwork reference and cidr to VirtualNetwork


rpc/common/moc_common_common.proto
```diff
@@ -358,3 +358,8 @@ enum CloudInitDataSource {
 	NoCloud = 0;
 	Azure = 1;
 }
+
+message ApiVersion {
```
sbgms, line 362: Minor: Ideally it should be Major.Minor.Build.Revision.

sbgms (dismissed):

rpc/common/moc_common_common.proto
nijosmsft: agree, i didnt want to add fields to protobuf that we are not using not.

sbgms (approved):

andkimsdn (approved):

PR #298, opened 2024-09-25
Error Codes for error propagation across agents

nikohobart:
## Changes Introduced by this PR
- Revamped errors pkg to use error codes and account for multierrors.
- Revised last error field on status to not include stacktrace and to include moc error codes.
- Existing methods maintain backwards compatibility.
- New/revised functions for returning grpc errors from server and receiving on client (GetGRPCError and ParseGRPCError). PRs to demonstrate how to use and successful MOC-CICD here: 
  - https://github.com/microsoft/moc-sdk-for-go/pull/275
  - https://github.com/microsoft/wssdcloudagent/pull/1129
  - https://github.com/microsoft/wssd-sdk-for-go/pull/295
  - https://github.com/microsoft/wssdagent/pull/1080
  - MOC-CICD Green Run: https://msazure.visualstudio.com/msk8s/_build/results?buildId=104990473&view=results

## Test Plan
- Extensive unit testing.
- Green runs on several repos using latest commit to show backwards compatibility:
  - Cloudagent: https://github.com/microsoft/wssdcloudagent/pull/1129
  - Nodeagent: https://github.com/microsoft/wssdagent/pull/1080
  - Cloud-operator: https://msazure.visualstudio.com/msk8s/_git/cloud-operator/pullrequest/10967626
  - Mgmtappl: https://msazure.visualstudio.com/msk8s/_git/mgmtappl/pullrequest/10967892
  - Moc Upgrade Sanity:
    - Ensures that store still works across moc upgrades (since we change the type of a last error field on status).


pkg/errors/codes/codes_string.go
```diff
@@ -0,0 +1,172 @@
...
+// Licensed under the Apache v2.0 license.
+package codes
+
+import (
+	"fmt"
+	"strconv"
+)
+
+// String returns the string representation of the MocCode
+func (c MocCode) String() string {
```
nikohobart, line 11: Use this in the old GetErrorCode() function for consistency.

pkg/errors/errors.go
```diff
@@ -8,59 +8,170 @@ import (
...
+// it will return the MocCode if all errors in the multierr match the same MocCode. If there are
+// multiple non-matching MocCodes, then it returns MocCode.MultipleErrors and false.
+//
+// GetMocErrorCode follows the following rules when parsing individual errors:
+//
+//   - If the error is nil, it returns MocCode.OK and true.
+//   - If the error is of type MocError, it returns the MocCode and true.
+//   - If the error has a Cause that is not nil and is of type MocError, it returns the MocCode of the Cause and true.
+//   - If both the error and its Cause are not of type MocError, it returns MocCode.Unknown and false.
+func GetMocErrorCode(err error) (moccodes.MocCode, bool) {
```
aeweston-msft, line 130: Nit: for my own understanding, what is the intent of the boolean?

aeweston-msft: Looks like it's used below in checkErrors but since one arg will always be a valid MOC error, isn't the equality check enough?

aeweston-msft (commented):
Mostly looks good, have a few clarifying questions and few things I thought can be removed, as discussed over chat

pkg/errors/errors.go
msftpgayam, line 130: It might be difficult to review these functions without seeing the usage in the agent code. I wonder if it makes sense to merge this PR after the agent PRs are reviewed as well.

pkg/errors/codes/codes_string.go
msftpgayam: Any idea where `GetErrorCode()` used?

pkg/errors/errors.go
```diff
@@ -210,6 +323,13 @@ func GetGRPCError(err error) error {
 	return err
 }
 
+// IsMocErrorCode wraps a call to GetMocErrorCode. It returns true only
+// if the error has the same MocCode as the given code (no string matching).
+func IsMocErrorCode(err error, code moccodes.MocCode) bool {
```
msftpgayam, line 328: Could you please point me to an example usage of this function?

rpc/common/moc_common_common.proto
```diff
@@ -60,7 +60,7 @@ enum HealthState {
 
 message Error {
 	string Message = 1;
-	int32 Code = 2;
+	uint32 Code = 2;
```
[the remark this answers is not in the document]
msftpgayam: I think this could cause issue during the upgrade. The status field may fail to unmarshal after the upgrade because the old status object is not compatible with the new object.

But given the code field is not used at all right now, I'm not sure how this would impact. I guess we need to add upgrade tests for this scenario.

pkg/errors/errors.go
```diff
@@ -364,18 +484,29 @@ func IsIPOutOfRange(err error) bool {
 	return checkError(err, IPOutOfRange)
 }
 
+func IsMultipleErrors(err error) bool {
```
msftpgayam, line 487: Could you please point me to an example usage of this function as well?

pkg/errors/errors.go
```diff
@@ -8,59 +8,170 @@ import (
...
+
+	moccodes "github.com/microsoft/moc/pkg/errors/codes"
+	"github.com/microsoft/moc/rpc/common"
 )
 
+// MocError is a trivial implementation of error that wraps the moc rpc Error struct.
+type MocError struct {
+	// None of MocError's fields should be modifiable after instantiation
+	// since errors are supposed to be immutable.
+	err *common.Error
```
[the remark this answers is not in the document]
msftpgayam: +1, I also would like to see the use cases of these functions.

pkg/errors/codes/codes_string.go
```diff
@@ -0,0 +1,172 @@
...
+	"GenericFailure":              GenericFailure,
+	"NoAuthenticationInformation": NoAuthenticationInformation,
+	"MeasurementUnitError":        MeasurementUnitError,
+	"QuotaViolation":              QuotaViolation,
+	"IPOutOfRange":                IPOutOfRange,
+	"MultipleErrors":              MultipleErrors,
+}
+
+// MocCodeFromString translates a string back to the corresponding MocCode constant
+func MocCodeFromString(s string) (MocCode, error) {
```
msftpgayam, line 167: Could you tell me where this is used?

pkg/errors/codes/codes.go
```diff
@@ -0,0 +1,160 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache v2.0 license.
+package codes
```
msftpgayam, line 3: Is there an advantage to putting codes in a separate package?

pkg/errors/errors.go
```diff
@@ -364,18 +484,29 @@ func IsIPOutOfRange(err error) bool {
...
 }
 
+func IsMultipleErrors(err error) bool {
+	return checkError(err, MultipleErrors)
+}
+
+// checkError checks if the wrappedError has the same MocCode as the err error according to GetMocErrorCode.
+// If the error is not matched by GetMocErrorCode and the error does not have a GRPC code (or is a GRPC Unknown code),
+// it will attempt to match the error strings through string matching (even for multierrors).
 func checkError(wrappedError, err error) bool {
```
msftpgayam, line 494: Can we write some unit tests for this function where the wrappedError not a `MocError`, instead it's an error constructed using errors package from golang with `errors.New()`? Because we cannot always be sure we are passing `MocError` to Is* functions.

rpc/common/moc_common_common.proto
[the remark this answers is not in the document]
nikohobart: Reason: mainly just because error codes, from my perspective, shouldn't be able to go negative.

Follow-up: yeah, this is persisted. My impression was that marshalling / unmarshalling wouldn't face any changes since we don't use the field (would only face issues if it was used with negative numbers). Since I'll be testing integrating this into 4 different repos independently, I think that would cover marshalling across different versions of status.

I can also include a run of the moc upgrade pipeline, which I think would cover compatibility with the store.

pkg/status/status.go
```diff
@@ -26,9 +28,16 @@ func InitStatus() *common.Status {
 // SetError
 func SetError(s *common.Status, err error) {
 	if err != nil {
-		s.LastError.Message = fmt.Sprintf("%+v", err)
+		// Remove the stack trace from the error message by calling Error() instead of fmt.Sprintf("%+v", err).
```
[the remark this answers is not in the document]
nikohobart: We still emit the stack trace to telemetry (if it is included on the error) when the error occurs. We just don't retain it on the last error field, which can be customer facing. (Prevents exposing internal code / making resource representation incredibly messy looking)

pkg/errors/codes/codes.go
```diff
@@ -0,0 +1,160 @@
...
+		return false
+	}
+
+	_, inMap := errorMessages[c]
+	if !inMap {
+		return false
+	}
+
+	s := c.String()
+	if strings.Contains(s, "MocCode(") {
```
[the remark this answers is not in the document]
nikohobart: Currently this function is only used inside the pkg to ensure that all string entries corresponding to an error code are filled out, was debating whether or not to expose this publicly.

I don't think there's a customer scenario in which the above two checks pass and this fails though, not unless we merge bypass on a failing unit test

pkg/errors/errors.go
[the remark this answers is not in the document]
nikohobart: Yeah, the divide is basically "default moc error code error" and "translating last error into a moc error code error without losing information".

So NewMocError for example helps create the base error codes (e.g., Not Found and its backwards-compatible message).

While NewMocErrorWithError is used to translate the last error struct across agents. By the time an error reaches cloudagent, it may have multiple layers, e.g., Msg: "The container 'x' was missing in location 'y': Not Found", Code:2. So we use NewMocErrorWithError to retain the message while recreating the moc error.

(Example of NewMocErrorWithError: https://github.com/microsoft/wssdcloudagent/blob/08bd4b8732526eec308e705f263f3c5bdd73add7/pkg/cloudagent/services/storage/virtualharddisk/wssd/wssd.go#L288)

pkg/errors/errors.go
nikohobart: It essentially is just a "was the error _actually_ Unknown or MultipleErrors? Or were we just unable to interpret the code / the error has multiple different codes?"

nikohobart: Fair point. Most of the functions are currently only used internally to the pkg. I think the ones I currently use in the agent PRs are:
1. NewMocErrorWithError: https://github.com/microsoft/wssdcloudagent/blob/08bd4b8732526eec308e705f263f3c5bdd73add7/pkg/cloudagent/services/storage/virtualharddisk/wssd/wssd.go#L288
2. IsMocErrorCode (for VHD decide to block detach or not)

pkg/errors/errors.go
nikohobart: https://github.com/microsoft/wssdcloudagent/blob/dccc5654228f3bd3f0ea4b647ba9998075c1152c/pkg/cloudagent/services/storage/virtualharddisk/virtualharddisk.go#L179

pkg/errors/errors.go
nikohobart: IsMultipleErrors doesn't currently have any usage. Mainly added since all errors have this functionality in the errors pkg.

pkg/errors/codes/codes_string.go
nikohobart: Will be removing after discussion with Anthony

pkg/errors/codes/codes.go
nikohobart: I view codes as the most base level error abstraction. Then the errors pkg as our wrapper for codes to provide enhanced functionality that integrates with the LastError field. Seemed to make sense to me in terms of not crowding the errors pkg and separation of concerns, but happy to hear other thoughts on this.

pkg/errors/errors.go
nikohobart: Yeah I have an existing unit test that covers this "TestCheckErrorBackwardsCompatible/Error_without_explicit_GRPC_Unknown_code_uses_string_matching". Do you think this suffices?

pkg/errors/codes/codes_string.go
nikohobart: Closing to help me organize PR revisions, feel free to re-open if I didn't address something.

pkg/status/status.go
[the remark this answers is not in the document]
nikohobart: Closing to help me organize PR revisions, feel free to re-open if I didn't address something.

pkg/errors/errors.go
nikohobart: Closing to help me organize PR revisions, feel free to re-open if I didn't address something.

pkg/errors/errors.go
nikohobart: Closing to help me organize PR revisions, feel free to re-open if I didn't address something.

pkg/errors/codes/codes.go
nikohobart: Closing to help me organize PR revisions, feel free to re-open if I didn't address something.

pkg/errors/errors.go
aeweston-msft: For ref the example usage is
` errors.IsMocErrorCode(err, codes.NotFound) `

What's the difference between this and IsNotFound?

pkg/status/status.go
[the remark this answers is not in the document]
aeweston-msft: Makes sense, thanks for changing that

pkg/errors/errors.go
nikohobart: IsNotFound is the old function, which will employ string matching. Whereas IsMocErrorCode only matches against the error code (won't match otherwise) in the root error.

rpc/common/moc_common_common.proto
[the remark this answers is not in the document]
aeweston-msft: From a practical standpoint, since we would never run out of int32 either, I would have left it as is to avoid the question of changing the format. It's not a big deal, so since its already changed its fine to leave as is IMO. 

For this specifically, since there is a question about the resources being loaded, could you please either do a manual test, or cross check that the upgrade pipeline actually validates that resources created before update are correctly loaded after. I would not assume that this specific case would actually cause the MOC upgrade pipeline to break. It might just ignore resources it couldn't load properly or something. I don't know what the behavior would be, so if the upgrade pipeline already covers this and you feel it is sufficient, please let me know. I am not familiar enough with the specifics of that pipeline and process to make a decision one way or another.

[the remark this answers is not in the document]
nikohobart: Yeah my big concern would've been some kind of resource corruption after we upgrade and start cloudagent/nodeagent back up, but there didn't seem to be any issues (most likely because we don't use the field, so it always defaults to 0) and the tests that could pass did.

The caveat here is that the upgrade is currently broken due to an unrelated regression that is causing a 100% failure rate on a particular test, so we're unable to see test progression past that.

pkg/errors/codes/codes.go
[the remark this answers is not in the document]
aeweston-msft: Not a big deal either way, but I would prefer to remove this from the business logic. If this is primarily aimed at validating the list is filled out correctly, I would suggest making that part of the unit test vs. putting a check here. Feel free to update or close as you see fit.

rpc/common/moc_common_common.proto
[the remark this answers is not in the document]
nikohobart: Double-checked. On cloudagent startup, we load the resource information from the store as plain interface{} (s). The first thing we do with this data is attempt a conversion to (*internal.Resource). If there were issues with this conversion, then cloudagent would just immediately crash since we don't perform it safely (we assume all the resources will be of type resource). So the upgrade pipeline (even though it didn't fully complete), would have caught this I'm fairly certain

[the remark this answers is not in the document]
aeweston-msft: Sounds good, thanks for checking.

aeweston-msft (commented):
LGTM for the most part, left a few minor comments that would IMO make the code more readable and result in fewer conversions between the different types (MocCode, MocError, error, and protobuf error). 

Discussed with Niko that further changes will be needed to account for converting gRPC errors to MocError, so not approving as more changes are planned.

apluchik:
Is there an MD file or doc for this?

rpc/common/moc_common_common.proto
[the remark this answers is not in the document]
apluchik: definitely would be good to run some upgrade testing. Would this be considered a breaking change between clients using older moc versions and agents using a newer one?

apluchik:
Changes look good from what I can tell. Having some documentation around the design of the change would help navigate the PR in the future

apluchik (approved):

rpc/common/moc_common_common.proto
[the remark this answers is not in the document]
nikohobart: It shouldn't be -- we've done extensive testing on different repos and pipelines. All original functions do retain their original functionality. We tested for backwards compatibility thru the moc-upgrade pipeline, along with adding this moc PR to different repos in the AKS HCI flow (e.g., cloud-operator, mgmtappl, cloudagent, etc.) and then making sure the automated pipelines are all green (even with all other repos using the old moc).

msftpgayam (dismissed):

rpc/common/moc_common_common.proto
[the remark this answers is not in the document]
apluchik: yep sorry forgot to close this, found this guide which shows this is a safe update:
https://protobuf.dev/programming-guides/proto3/#updating

apluchik (approved):

msftpgayam (approved):

PR #302, opened 2024-10-09
Add error message redaction functionality to /pkg/redact

pathuang-sia:
Redaction to error message in addition to the resource properties is necessary, as illustrated by this bug: https://msazure.visualstudio.com/One/_workitems/edit/29734266?view=fullscreen

TODO:
- Add unit test for error redaction functionality
- validation
- code cleanup for readability and performance
- further refinement to see what kind of sensitive field value we should redact. (e.g., we don't want to redact random numbers or words in the error message just because they are sensitive. They might not even be relevant to the error message)

### DIFF
--- a/pkg/redact/redact.go
+++ b/pkg/redact/redact.go
@@ -48,6 +48,70 @@ func Redact(msg interface{}, val reflect.Value) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pathuang-sia:
duplicate of this PR: https://github.com/microsoft/moc/pull/303/files, closing

PR #303, opened 2024-10-14
moc repo - redact error - bug: 29734266 - GalleryImage SAS URI not redacted in Telemetry

rbalwani:
Customer's failed galleryimage creation errored out initially at the node agent level, which should have redacted when showing the error message.

The redaction however is not redacting the error message at the node agent level because current redaction only operates on the virtualharddisk properties that are marked as sensitive in the protobufs. The proxy error however, is a common container error that prints out the uri proxy by default; the error message itself does not go through any redacting. causing the SAS URI to propagate (as part of the error message) all the way up to the azstackhci-operator level

This PR adds redact of error message


pkg/validations/proxy_validation_test.go
```diff
@@ -28,7 +26,7 @@ func Test_ValidateProxyURL(t *testing.T) {
 	}
 }
 
-func Test_TestProxyUrlConnection(t *testing.T) {
+/*func Test_TestProxyUrlConnection(t *testing.T) {
```
pathuang-sia, line 29: Is this change related?

pathuang-sia (approved):

pkg/validations/proxy_validation_test.go
rbalwani: reverted

pathuang-sia (approved):

pkg/redact/redact.go
```diff
@@ -48,6 +49,36 @@ func Redact(msg interface{}, val reflect.Value) {
 	}
 }
 
+// Redact - removes data from fields marked as sensitive given a proto message struct
```
friggitello6, line 52: NIT: update comments to reflect right function name and intent.  Same for the below two functions.

pkg/redact/redact.go
```diff
@@ -48,6 +49,36 @@ func Redact(msg interface{}, val reflect.Value) {
...
+}
+
+// RedactedMessage - returns a copy of a proto message struct with data from fields marked as sensitive redacted
+func RedactedError(msg interface{}, err *error) interface{} {
+	rMsg := proto.Clone((msg).(proto.Message))
+	RedactSensitiveError(rMsg, reflect.ValueOf(rMsg), err)
+	return rMsg
+}
+
 func redactMessage(msg interface{}, val reflect.Value) {
```
friggitello6, line 82: NIT: Although we did quite a bit of code refactoring, it only applies to redacting sensitive information from error messages. The implementation is very similar to redacting sensitive value from the field, but we have duplicated code. We should be able to reuse it with some form of default values, such as:

RedactMessageWithDefault(msg) {
    RedactMessage(msg, nil)
}
Given the urgency, we could just file an item to track future improvements for refactoring.

friggitello6 (dismissed):

pkg/redact/redact.go
```diff
@@ -48,6 +49,36 @@ func Redact(msg interface{}, val reflect.Value) {
...
+		// TODO: Implement map logic. Currently only certificates in identity use it, and are redacted
+	case reflect.Ptr:
+		RedactSensitiveError(msg, val.Elem(), err)
+	case reflect.Struct:
+		redactErrorMessage(msg, val, err)
+	}
+}
+
+// RedactedMessage - returns a copy of a proto message struct with data from fields marked as sensitive redacted
+func RedactedError(msg interface{}, err *error) interface{} {
```
friggitello6, line 76: This function name is very misleading. It redacts sensitive data from err and doesn't modify msg. Why do we return the cloned message? Why do we even need to clone msg since we are not modifying it? The function name gives the impression that the returned value is an error message with sensitive data redacted, but we are just doing in-place redacting.

pkg/redact/redact.go
rbalwani: added/updated comments

pkg/redact/redact.go
rbalwani: Will add future work item for it

pkg/redact/redact.go
rbalwani: yeah, I removed clone and return statement. I was about to add a review for it, thanks for pointing that out.

pkg/redact/redact.go
```diff
@@ -48,6 +49,37 @@ func Redact(msg interface{}, val reflect.Value) {
...
+		RedactSensitiveError(msg, val.Elem(), err)
+	case reflect.Struct:
+		redactErrorMessage(msg, val, err)
+	}
+}
+
+// RedactedError redacts sensitive information from the provided error message.
+// It takes a message of any type and a pointer to an error, and processes the message
+// to remove any sensitive data from error message.
+func RedactedError(msg interface{}, err *error) {
```
friggitello6, line 79: Thought I added following comment, but couldn't find it.  Adding here again.

This function name is very misleading. It redacts sensitive data from err and doesn't modify msg. Why do we return the cloned message? Why do we even need to clone msg since we are not modifying it? The function name gives the impression that the returned value is an error message with sensitive data redacted, but we are just doing in-place redacting.

rbalwani: I removed clone message yesterday

friggitello6 (approved):

pathuang-sia (approved):

PR #307, opened 2024-10-16
[Zone] Storage Affinity

bgriddaluru:
## Background

1. For split rack cluster setup, ECE action plan will create workload volumes, one per node, and affinity of volumes to nodes and zones is set using preferred owners.

## Problem

1. If VM is associated with a zone., currently container is picked in a round robin fashion. 
2. If volume backing the container is active on another node/zone, then IO performance will be degraded

## Proposed solution

1. For container, read the preferred owner from failover cluster and store the info in container spec
2. During creation of a VM which is associated with zone, try to pick a container in following order

### order:
1. Container with same preferred owner
2. Container's preferred owner is in same zone
3. Round-Robin of any available containers

#### Moc
1.  Add 'PreferredOwner' field in Container Info message

#### Moc-Pkg
1.  Add wrapper function for WMI query to fetch preferred owners for the cluster shared volume from the csv's cluster group

#### Node-Agent
1. Add new provider type ' Failover cluster' , which is wrapper on top of default local provider. 
2. This additionally fetches the preferred owner for container from failover cluster layer and updates in the container info spec.

#### Cloud-Agent
1. During VM creation, if a VM is associated with Zone, then select a container that has same preferred owner or preferred owner is in same zone.


rpc/common/moc_common_storageinfo.proto
```diff
@@ -11,6 +11,7 @@ import "moc_common_common.proto";
 message StorageContainerInfo {
 	Capacity capacity  = 1;
 	Latency latency	= 2;
+	string preferredOwner = 3;
```
piyushka17, line 14: Can you provide more details on the property? Is this node/zone?

piyushka17, line 14: If this applies for MOC cloudagent interface, then it should be reflecting zone.
Also the name should not sound specific to failover cluster.

bgriddaluru: It is 'preferred Owner node'. I will change it accordingly after further disucssion.

bgriddaluru: 1. In VM creation flow, preferring container which is on same node first (determined by most preferred owner). If not, preferring containers in same zone. 
2. So, we need to maintain 'most preferred node/owner' for container and in cloud agent zone information is derived from node during container selection.
3.  If requirement is to expose 'zone' name in container spec, then I can either derive it from preferred owner in cloud agent or fetch it from volume's custom property set by ECE action plan on node agent side . Please let me know the preferred approach.

madhanrm: What is the intent of this property.

Is this input or readonly?

bgriddaluru: Yes, this is a read only field for user and populated by MOC logic.

Similar to fields like capacity, this info is filled in node agent layer from Failover Cluster and then sent to cloud agent in the Create/Get response. 

Based on feedback from Piyush, I am also adding zone name. However, for the best IO performance, we still need to pick a container that is active in the same node (and zone) if available.  So additional field of 'node name'(preferred owner) is also required.

Workflow
1. ECE action plan creates workload CSVs (one per node). Sets Affinity for each CSV to a Zone.  For CSVs in a particular zone, the preferred owners are changed such that each node is a preferred owner to a specific CSV.
2.  When container is created, preferred owner info is fetched from backing CSV in node agent and sent to cloud agent.
3. Node agent is not zone aware, so only node info is sent to cloud agent.
4. In cloud agent during VM creation, this node name / preferred owner field is to used to pick a container that is active same node or zone as VM for better IO performance.

rpc/common/moc_common_storageinfo.proto
```diff
@@ -11,6 +11,8 @@ import "moc_common_common.proto";
 message StorageContainerInfo {
 	Capacity capacity  = 1;
 	Latency latency	= 2;
+	string nodeName = 3;
+	string zoneName = 4;
```
piyushka17, line 15: zoneName -> zone? (to be consistent with VM, PlacementGroup).

Same for node.

bgriddaluru: Fixed in new patch.

piyushka17 (approved):

rpc/common/moc_common_storageinfo.proto
bgriddaluru: Discussed offline. I will update cloud agent logic to populate the zone name in the container info.

mkatturu (approved):

PR #309, opened 2024-10-16
cherry-pick - bug: 29734266 - GalleryImage SAS URI not redacted in Telemetry

rbalwani:
cherry-pick - moc repo - redact error - bug: 29734266 - GalleryImage SAS URI not redacted in Telemetry.

Customer's failed galleryimage creation errored out initially at the node agent level, which should have redacted when showing the error message.

The redaction however is not redacting the error message at the node agent level because current redaction only operates on the virtualharddisk properties that are marked as sensitive in the protobufs. The proxy error however, is a common container error that prints out the uri proxy by default; the error message itself does not go through any redacting. causing the SAS URI to propogate (as part of the error message) all the way up to the azstackhci-operator level:

### DIFF
--- a/pkg/redact/redact.go
+++ b/pkg/redact/redact.go
@@ -5,6 +5,7 @@ package redact
 
 import (
 	"encoding/json"
+	"fmt"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


friggitello6 (approved):

bigjeffxia2022 (approved):
Sounds good.

PR #310, opened 2024-10-17
added priority field to VM in RPC

connievi:


rpc/nodeagent/compute/virtualmachine/moc_nodeagent_virtualmachine.proto
```diff
@@ -139,6 +139,7 @@ message VirtualMachine {
 	GuestAgentConfiguration guestAgent = 15;
 	VirtualMachineAgentInstanceView guestAgentInstanceView = 16;
 	ZoneConfiguration zoneConfiguration = 17;
+	Priority priority = 18;
```
bgriddaluru, line 142: Items for discussion. Please check with team.

1. 'failoverPriority' is more user friendly compared to 'priotiry'
2. Should we define a message called 'failover properties' and put the field in it ? So that in future new failover related fields can be added under it instead of VM spec

connievi: During the design review with MOC team, Madhan and others said that they wanted the field to be `priority` to make it broader and not limited to failover for future purposes

piyushka17:
Diff looks fine, waiting for a confirmation on validation/regression check.

PR #237, opened 2024-02-14
VM, NIC and disk hydration

neerajdixit-msft2:
This PR is part of the VM, NIC and disk hydration feature that spans the whole HCI stack:

- Design:
  - https://microsoft-my.sharepoint.com/:w:/p/ndixit/EXvBh8LSdEFLkSbKHmNZiCIBqEJtgiMlkHeoNBiybXH4FA?e=jsqS2J
- ADO feature
  - https://msazure.visualstudio.com/One/_workitems/edit/28572249
- Testing
  - Testing has been executed iteratively as part of the Trailblazer scrum feature delivery. We are tracking testing at https://microsoft-my.sharepoint.com/:w:/p/ndixit/EcEmFhNIOQZOui-qnJu_kxIBMrPD6K8nTg-5Upu0bEm7CQ?e=0F6u6C 

---

These changes include the protobuf changes required to create a new operation `HYDRATE`

PR #318, opened 2024-11-05
Added support for Upload Disk Feature

damanMSFT:


pkg/status/status.go
```diff
@@ -125,6 +126,26 @@ func GetPlacementStatus(s *common.Status) common.PlacementStatusType {
 	return s.GetPlacementStatus().GetStatus()
 }
 
+// Set UploadError
+func SetUploadError(s *common.Status, err ...error) {
```
niontive, line 130: Why is this exported? It looks like it's only used by this package.

pkg/status/status.go
```diff
@@ -125,6 +126,26 @@ func GetPlacementStatus(s *common.Status) common.PlacementStatusType {
...
+	s.UploadStatus.LastUploadError = errors.ErrorToProto(err[0])
+}
+
+// SetUploadStatus
+func SetUploadStatus(s *common.Status, dProgressPercentage, dUploadSizeInBytes, dFileSizeInBytes int64, err ...error) {
```
niontive, line 135: See above comment about export.

Out of scope for this PR - I'm curious why this file has `func Set...()` and passes in a `common.Status` - it'd make sense for the `Status` object to have methods to set its upload status, upload error, etc.

pkg/status/status.go
```diff
@@ -146,6 +167,8 @@ func GetStatuses(status *common.Status) map[string]*string {
 	if placementStatus != "" {
 		statuses["PlacementStatus"] = &placementStatus
 	}
+	ustate := status.GetUploadStatus().String()
```
blairurish, line 170: Just to keep it consistent with the changes being made in the hydration PR, could we null check the result of status.GetUploadStatus like what's done here: https://github.com/microsoft/moc/pull/237/files#diff-1deea3647518183ee4915dd3c10ccfc7bb60e1ed860673c99eefaf5cfeefaa6b

pkg/status/status.go
damanMSFT: Yeah I agree. This could be refactored to something like ```func (s *Status) SetUploadStatus()```. I am not sure why this is done this way though. I just followed the pattern I saw in the rest of the file.

pkg/status/status.go
damanMSFT: Done!

pkg/status/status.go
```diff
@@ -129,6 +130,26 @@ func GetPlacementStatus(s *common.Status) common.PlacementStatusType {
 	return s.GetPlacementStatus().GetStatus()
 }
 
+// Set UploadError
+func SetUploadError(s *common.Status, err ...error) {
+	s.UploadStatus.LastUploadError = errors.ErrorToProto(err[0])
```
jocardeMSFT, line 135: I realize this might be following a pattern but there is a flaw that can result in a runtime panic.  If someone calls this function with 0 errors (err ...error) is a variadic paramater that can be 0..many errors.

if it is 0 errors

this function will result in a runtime panic -> err[0] will panic due to out of bounds access.

suggested fix:

func SetUploadError(s *common.Status, err ...error) {
    if len(err) == 0 {
        s.UploadStatus.LastUploadError = nil
        return
    }
    s.UploadStatus.LastUploadError = errors.ErrorToProto(err[0])
}

jocardeMSFT, line 135: I realize this might be following a pattern but there is a flaw that can result in a runtime panic.  If someone calls this function with 0 errors (err ...error) is a variadic paramater that can be 0..many errors.

if it is 0 errors

this function will result in a runtime panic -> err[0] will panic due to out of bounds access.

suggested fix:

func SetUploadError(s *common.Status, err ...error) {
    if len(err) == 0 {
        s.UploadStatus.LastUploadError = nil
        return
    }
    s.UploadStatus.LastUploadError = errors.ErrorToProto(err[0])
}

jocardeMSFT: If the only place we are calling is below then why do we need a variadic parameter?

rpc/common/moc_common_common.proto
```diff
@@ -124,6 +125,13 @@ message PlacementStatus  {
...
 }
 
+message UploadStatus {
+	int64 progressPercentage = 1;
+	int64 uploadSizeInBytes = 2;
+	int64 fileSizeInBytes = 3;
+	Error lastUploadError = 4;
+}
+
 message Status {
```
jocardeMSFT, line 135: I feel like it wouldn't be a bad idea to throw a todo here to re-evaluate how to approach resource Status so we don't just keep creating a giant status that gets applied to every resource.  Like a VM with a nil upload status.

jocardeMSFT (commented):
Completed inital pass review

pkg/status/status.go
damanMSFT: Thanks for pointing it out. I actually fixed it by removing the variadic err param.

rpc/common/moc_common_common.proto
damanMSFT: Agreed. Added a comment!

jocardeMSFT (approved):

rpc/common/moc_common_common.proto
```diff
@@ -346,6 +356,7 @@ enum ProviderAccessOperation {
 
 	VirtualHardDisk_Create          = 3100;
 	VirtualHardDisk_Update          = 3101;
+	VirtualHardDisk_Upload 			= 3102;
```
rbalwani, line 359: nit please remove tabs for formatting

rbalwani (dismissed):

rpc/common/moc_common_common.proto
damanMSFT: I am not sure why it shows up like this here. It looks good on my local setup.

damanMSFT: image

damanMSFT: Done

rbalwani (approved):

jocardeMSFT (approved):

PR #319, opened 2024-11-08
removed redundancy in makefile

jocardeMSFT:
Self explanatory, simple housekeeping fix


Makefile
```diff
@@ -14,7 +14,7 @@ export GO111MODULE=on
 #
 PKG := 
 
-all: format test unittest
+all: format test
```
mduppre, line 17: While you're at it, I don't think there is any need to have `unittest` at all, is there? Let's just remove `test: unittest` and rename the `unittest` target to `test`?

mduppre (dismissed):

Makefile
mduppre: Sure, I generally don't think it's worth leaving placeholders for future changes to come (which often never come), but I don't feel strongly at all about this tiny instance :-)

blairurish (dismissed):

blairurish (approved):

halleremash (approved):

PR #332, opened 2025-01-31
Redact sensitive json field from nodeagent vhd

Arkanayan:


pkg/redact/redact.go
```diff
@@ -150,12 +152,15 @@ func redactJsonSensitiveField(val reflect.Value) {
 	if err := json.Unmarshal([]byte(validJsonString), &jsonData); err != nil {
 		return
 	}
-	for key := range jsonData {
-		// This can be extended to an array of sensitive keys if needed
-		if key == "private-key" {
-			jsonData[key] = RedactedString
+
+	sensitiveKeys := [...]string{"private-key", "sasURI"}
```
raghavendra-nataraj, line 156: Is the SAS url being persisted today?

liwonMS (approved):

pkg/redact/redact.go
```diff
@@ -253,12 +258,19 @@ func redactErrorJsonSensitiveField(val reflect.Value, errMessage *error) {
 	if err := json.Unmarshal([]byte(validJsonString), &jsonData); err != nil {
 		return
 	}
-	for key := range jsonData {
-		// This can be extended to an array of sensitive keys if needed
-		if key == "private-key" {
-			if strVal, ok := jsonData[key].(string); ok && errMessage != nil && *errMessage != nil && strVal != "" {
-				redactSensitiveField(strVal, errMessage)
-			}
+	sensitiveKeys := [...]string{"private-key", "sasURI"}
```
friggitello6, line 261: Are we going to miss others if we just whitelist these two in common lib like MOC?

liwonMS: Actually the original function, if you look at it, it only work on Private-key, and all others fields are no op.

pkg/redact/redact_test.go
```diff
@@ -193,3 +196,53 @@ func TestRedactErrorJsonSensitiveField(t *testing.T) {
...
+	tests := []struct {
+		name      string
+		args      args
+		uri       string
+		wantError bool
+	}{
+		{name: "valid URI", args: struct {
+			err error
+			uri string
+		}{err: &url.Error{
```
jocardeMSFT, line 214: nit: why use an anonymous struct type for the args value here when you have the args struct defined above?  ...

        {
            name: "valid URI",
            args: args{
                err: &url.Error{
                    Op:  "Head",
                    URL: "https://sas.azure.net/test?sp=sljsdf&st=2025-01-31T10:33:25Z&sv=2022-10-02&spr=3lskdjfoi23y9owh9u23fgn",
                    Err: errors.New("unable to reach host"),
                },
                uri: "https://sas.azure.net/test?sp=sljsdf&st=2025-01-31T10:33:25Z&sv=2022-10-02&spr=3lskdjfoi23y9owh9u23fgn",
            },
            wantError: false,
        },

pkg/redact/redact_test.go
```diff
@@ -193,3 +196,53 @@ func TestRedactErrorJsonSensitiveField(t *testing.T) {
...
+
+func TestRedactErrorURL(t *testing.T) {
+	type args struct {
+		err error
+		uri string
+	}
+	tests := []struct {
+		name      string
+		args      args
+		uri       string
```
jocardeMSFT, line 208: tt.uri is not used in the test, remove.

pkg/redact/redact_test.go
```diff
@@ -193,3 +196,53 @@ func TestRedactErrorJsonSensitiveField(t *testing.T) {
...
+			URL: "",
+			Err: errors.New("unable to reach host"),
+		},
+			uri: ""}, wantError: true},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			RedactErrorURL(tt.args.err)
+			if tt.wantError != strings.Contains(tt.args.err.Error(), tt.args.uri) {
+				t.Errorf("RedactErrorURL() got = %v, want %v", tt.args.err, tt.wantError)
```
jocardeMSFT, line 244: nit: The message we are sending in this error message seems... unintuitive ?  for example if the uri remains when we want it gone we will print

RedactErrorURL() got = **error message including URI I assume**, want false

???

rpc/nodeagent/storage/virtualharddisk/moc_nodeagent_virtualharddisk.proto
```diff
@@ -67,7 +69,7 @@ message AzureGalleryImageProperties {
 message VirtualHardDisk {
 	string name = 1;
 	string id = 2;
-	string source = 3 [(sensitive) = true];
+	string source = 3 [(sensitivejson) = true];
```
jocardeMSFT, line 72: for my learning, what is the diff between sensitivejson and sensitive?

pkg/redact/redact.go
jocardeMSFT: My immediate thought is if the error is wrapped, are we removing the URL from all error messages enclosed or just top level?

liwonMS:
we need to do the same for senstivie field for cloudagnet galleryimage.

pkg/redact/redact.go
rbalwani: This can be improved later. Currently we are only redacting error message from Virtalharddiskdrive which only has source labelled sensitivejson. 

JSON object is AzureGalleryImageProperties which has sasURI as field which is marked sensitive.

rpc/nodeagent/storage/virtualharddisk/moc_nodeagent_virtualharddisk.proto
rbalwani: Here AzureGalleryImageProperties struct is encoded as JSON string.

Within AzureGalleryImageProperties struct sasURI is labelled as sensitive.

rpc/nodeagent/storage/moc_nodeagent_virtualharddisk_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+func TestRedactedError_VHD(t *testing.T) {
+	// Has issue with ampersand (&); queryescape is needed to pass
+	uri := url.QueryEscape(`https://test.blob.core.windowssdfsdf.net/testvhd/test.vhdx?sp=r&st=2025-01-31T10:33:25Z&se=2025-01-31T18:33:25Z&spr=https&sv=2022-11-02&sr=b&sig=dfdf`)
+
+	azureProp := AzureGalleryImageProperties{SasURI: uri}
+	propertiesJson, e := json.Marshal(azureProp)
+	if e != nil {
+		t.Error(e)
+	}
```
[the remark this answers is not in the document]
rbalwani: can we fixed later

rbalwani (approved):

PR #339, opened 2025-03-04
Update key structures to support KMSv2

jirenugo:
[Feature docs](https://github.com/microsoft/wssdagent/blob/d8092c22126b85cf279ee81f9040fe796b2f075b/doc/features/security/KMSv2.md)


rpc/cloudagent/security/keyvault/key/moc_cloudagent_key.proto
```diff
@@ -34,6 +35,7 @@ message KeyOperationResponse {
 	string Data = 1 [(sensitive) = true];
 	google.protobuf.BoolValue Result = 2;
 	string  Error = 3;
+	Key key = 4;
```
raghavendra-nataraj, line 38: Is this the private key we are passing back?

raghavendra-nataraj: We should ensure the private key is not going out of MOC

raghavendra-nataraj (changes requested):

rpc/cloudagent/security/keyvault/key/moc_cloudagent_key.proto
jirenugo: It's the same structure as the one returned by `Get` which does not include the private key. We really only need the `keyAge` and the `keyVersion`. Let me modify the structure to return just the fields necessary.

raghavendra-nataraj (approved):

raghavendra-nataraj (dismissed):

raghavendra-nataraj (approved):

madhanrm (approved):

PR #344, opened 2025-04-01
add NO_KEY_WRAP

danmamsft:
This is the 1st PR of below work item "Import private key in plain text":
https://msazure.visualstudio.com/One/_workitems/edit/31918085

NO_KEY_WRAP is already available in moc-sdk-for-go. For this work item, I'm using NO_KEY_WRAP to identify when to import a plaintext key


SethBe (approved):
:shipit:

raghavendra-nataraj (approved):

PR #349, opened 2025-04-15
Update to jwt/v4 v4.5.2

damanMSFT:
This is done to fix the Component Governance Alert here https://msazure.visualstudio.com/msk8s/_componentGovernance/moc-operator/alert/12823868?typeId=26363705


msftpgayam (approved):

pkg/auth/auth_test.go
```diff
@@ -18,7 +18,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/golang-jwt/jwt"
+	"github.com/golang-jwt/jwt/v4"
```
raghavendra-nataraj, line 21: IS this backward compatible to older jwt version?

damanMSFT: I believe so. I looked it up here [jwt package - github.com/golang-jwt/jwt/v4 - Go Packages](https://pkg.go.dev/github.com/golang-jwt/jwt/v4#section-readme)

image

raghavendra-nataraj (approved):

PR #361, opened 2025-05-29
add support for token authentication between cloud and node agent

wecha-ms-v2:
This PR adds a new poptokenauth component that generates the token to be passed to the node agent. For now it returns a fake token


seanyen (approved):
LGTM

PR #365, opened 2025-06-17
[CMP] Add poptoken creation and validation to moc 

wecha-ms-v2:


pkg/auth/poptoken/shrpoptoken.go
```diff
@@ -0,0 +1,206 @@
...
+	bs := make([]byte, 4)
+	binary.BigEndian.PutUint32(bs, uint32(e))
+
+	bs = bs[1:] // drop most significant byte - leaving least-significant 3-bytes
+	ss := base64.URLEncoding.EncodeToString(bs)
+	return ss
+}
+
+// Append custom claims to the existing ShrPopTokenBody.
+func (pop *ShrPopToken) appendCustomClaimsToBody(customClaims map[string]interface{}) map[string]interface{} {
```
keystroke, line 119: why would claims be added to body? once created the token should not be edited. you can store "pending" claims in a map from the beginning and not have to martial into one here. But ideally you just take such claims in the interface in the first place.

calling for SHR pop should look something like this (this directly inits the struct but can also expose a "ctor function" instead of externally init the struct and handle the optional inclusions of nonce query and custom claims which you can coalesce into defaults):

```golang
authScheme := &shrpopauth.ShrPoPAuthScheme{
    Method: "GET",
    Host:   "graph.microsoft.com",
    Path:   "/v1.0/me",
    Query:  "", // optional
    Nonce:  nonce, // optional, if null random one generated and used
    Claims: map[string]interface{}{ // optional
	"custom_claim": "value",
    }
}
```

when you construct the body is when the token is being formatted so you should only ever need to construct it once, so appending to it doesn't make sense

pkg/auth/poptoken/rsakeymanager.go
```diff
@@ -0,0 +1,120 @@
...
+	RsaSize = 2048
+	Kty     = "RSA"
+	Alg     = "RS256"
+)
+
+func generatePrivateKey() (*rsa.PrivateKey, error) {
+	return rsa.GenerateKey(rand.Reader, RsaSize)
+}
+
+// Background job to continuously refresh the keypair in a best effort basis.
```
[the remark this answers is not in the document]
wecha-ms-v2: that is the implementation. We pass a keypair to the func NewNodeAgentPopTokenAuthScheme "ctor()" which is cached until the token is generated. [1]. From NewNodeAgentPopTokenAuthScheme perspective, it is not concern about rotation of keypair.

the rsakeymanager implementation is an independent singleton that focuses on rotating the keypair in the background. 

We can remove rsakeymanager and simplify the logic if we generate a new keypair per NewNodeAgentPopTokenAuthScheme initialization, but I don't think that is efficient. 

[1] https://github.com/microsoft/moc/blob/user/wecha/pop2/pkg/auth/poptoken/nodeagentpoptokenscheme.go#L56

pkg/auth/poptoken/shrpoptoken.go
```diff
@@ -0,0 +1,206 @@
...
+
+// https://datatracker.ietf.org/doc/html/draft-ietf-oauth-signed-http-request-03#section-3
+type ShrPopTokenBody struct {
+	Cnf Cnf `json:"cnf"`
+	// timestamp
+	Ts int64 `json:"ts"`
+	// access token
+	At string `json:"at"`
+	// random unique value to prevent replay attack. not used
+	NonCe string `json:"nonce"`
```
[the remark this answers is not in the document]
wecha-ms-v2: mistype? nonce is not implemented at this point, no plans to at this point.

pkg/auth/poptoken/shrpoptoken.go
wecha-ms-v2: yes we add the claims to the token while is generated.  When GenerateToken() call [1, it calls the private function appendCustomClaims() to add claims. This is an internal implementation detail,  I can move the logic in appendCustomClaims() back into GenerateToken() if it helps.

nodeagentpoptokenscheme is the "interface" you are thinking of as it implements the MSAL callback. See [2] shrpoptoken is more of the private class that does the heavy lifting. 

[1] https://github.com/microsoft/moc/blob/user/wecha/pop2/pkg/auth/poptoken/shrpoptoken.go#L140
[2] https://github.com/microsoft/moc/blob/user/wecha/pop2/pkg/auth/poptoken/nodeagentpoptokenscheme.go#L56

pkg/auth/poptoken/shrpoptoken.go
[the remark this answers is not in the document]
keystroke: Nonce is critical for protection against replay attacks, hence the comment and the note on client-side generation we discussed with idx team (which covers this outside of the pre-generation attack which is more esoteric and less relevant here). This does not need to be the full www-authenticate header with signed payload interaction, but just a client-side guid generated with each call. The server-side tracking and rejection can come "later" but is also pretty easy to implement.

pkg/auth/poptoken/rsakeymanager.go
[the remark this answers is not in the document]
keystroke: Yes let's keep it simple and not expose keymgmt to callers (e.g. do not make the cached key on the pop token configuration public, or as a param to the ctor function.

Instead, just have it use internal cached key, and jit generate a new one as needed. That is, we do not generate a new key for every call; we only generate a new key based on a lifetime calculation.

```
func NewPoPAuthScheme(method, host, path, query, nonce string) *PoPAuthScheme {
	scheme := &PoPAuthScheme{
		Method: method,
		Host:   host,
		Path:   path,
		Query:  query,
		Nonce:  nonce,
	}

	popKeyMutex.Lock()
	defer popKeyMutex.Unlock()

	now := time.Now()
	if popKeyActive == nil || now.Sub(popKeyActive.createdAt) > popKeyLifetime {
		privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
		if err != nil {
			panic(fmt.Sprintf("popauth: failed to generate PoP key: %v", err))
		}

		jwk := buildJWK(privateKey.Public())
		jwkBytes, _ := json.Marshal(jwk)
		jwkJson := string(jwkBytes)

		kid := computeKeyID(jwkBytes)
		reqCnf := computeReqCnf(kid)

		popKeyActive = &popKey{
			privateKey: privateKey,
			createdAt:  now,
			reqCnf:     reqCnf,
			kid:        kid,
			jwkJson:    jwkJson,
			jwk:        jwk,
		}
	}

	scheme.key = popKeyActive
	return scheme
}
```

Above, each time the caller goes to make a new pop auth scheme, we will either continue using the same pop key we have, or generate a new one if the old one needs to be rotated. This can just be done inline.

pkg/auth/poptoken/msalauthprovider.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil, errors.Wrapf(err, "failed to create confidential client")
+	}
+
+	return &confidentialClient, nil
+}
+
+func (m MsalAuthProvider) GetToken(targetResourceId string) (string, error) {
+
+	// TODO: the underlying client certificate will be refreshed, hence we need to also pick up the new certificate
+	// Longer run we can cache the client but for now we will refresh the client for every token call.
```
keystroke, line 54: invert this - add the "client hydration" later, rather than starting out with reading cert from disk for every get token call

pkg/auth/poptoken/nodeagentpoptokenscheme.go
```diff
@@ -0,0 +1,66 @@
...
+	return a.shrPopToken.GenerateToken(accessToken, time.Now(), customClaims)
+}
+
+// Return the token type. Must be "pop"
+func (a *NodeAgentPopTokenAuthScheme) AccessTokenType() string {
+	return a.shrPopToken.Header.Typ
+}
+
+// Create a new instance of NodeAgentPopTokenAuthScheme. Pass in the custom claims to be set in the pop token here, e.g. resourceId
+func NewNodeAgentPopTokenAuthScheme(resourceId string, rsaKeyPair *RsaKeyPair) (*NodeAgentPopTokenAuthScheme, error) {
```
keystroke, line 56: the scheme can be more generic, let the provider thing be the one with context on the specific claims going into it. this ensures proper separation of concerns and flexibility in implementing specific claim flows.

e.g., the poptokenauthscheme type could just accept a dictionary of claims for the wrapper pop token. its the caller that add the context for fields from dpop or for a new "grpc pop" where we would come up with scoping for grpc claims like method name etc.

```golang
func NewPopTokenAuthScheme(customClaims map[string]interface{}) *PopTokenAuthScheme
```
thus the same scheme type here can used by caller to add nonce, dpop claims, or grpc ones. could even be used as-is to call public graph API like in sample I shared, giving another test avenue to ensure the pop tokens are created correctly.

```golang
dPopScheme := NewPopTokenAuthScheme(map[string]interface{}{
	"htm": "GET",
	"htu": "https://api.example.com/resource",
	"jti": uuid.NewString(),
})

grpcPopScheme := NewPopTokenAuthScheme(map[string]interface{}{
	"rpc": "/my.service.Foo/Bar",
	"svc": "inventory",
	"trace": map[string]interface{}{
		"id":  "abc123",
		"seq": 42,
	},
})
```

one can then create wrapper ctor functions for schematized parameter passing:

```golang
func NewSHRPopScheme(method, host, path, query, nonce string) *PopTokenAuthScheme {
	claims := map[string]interface{}{
		"m": method,
		"u": host,
		"p": path,
	}
	if query != "" {
		qHash := sha256.Sum256([]byte(query))
		claims["q"] = []interface{}{
			extractQueryParamNames(query),
			base64.RawURLEncoding.EncodeToString(qHash[:]),
		}
	}
	if nonce != "" {
		claims["nonce"] = nonce
	}
	return NewPopTokenAuthScheme(claims)
}

func NewDPoPScheme(method, url, jti string) *PopTokenAuthScheme {
	if jti == "" {
		jti = uuid.NewString()
	}
	return NewPopTokenAuthScheme(map[string]interface{}{
		"htm": method,
		"htu": url,
		"jti": jti,
	})
}

func NewGRPCPopScheme(fullMethod string, service string, traceID string, seq int) *PopTokenAuthScheme {
	claims := map[string]interface{}{
		"rpc": fullMethod,
		"svc": service,
	}
	if traceID != "" {
		claims["trace"] = map[string]interface{}{
			"id":  traceID,
			"seq": seq,
		}
	}
	return NewPopTokenAuthScheme(claims)
}
```

pkg/auth/poptoken/nodeagentpoptokenscheme.go
```diff
@@ -0,0 +1,66 @@
...
+	keyPair     *RsaKeyPair
+	ResourceId  string
+}
+
+// Return the claim containg the pop token kid that will be added to the Entra access token.
+func (a *NodeAgentPopTokenAuthScheme) TokenRequestParams() map[string]string {
+
+	reqCnfBase64, err := a.shrPopToken.GetReqCnf()
+	if err != nil {
+		return map[string]string{}
```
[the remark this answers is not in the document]
wecha-ms-v2: honestly, that should be impossible to fail at this point, since we "calculate" the cnf at ctor() time. better to remove the err check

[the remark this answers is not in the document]
keystroke: the error is returned by lower level function, if such a low-level or pathological error occurs should propagate it instead if ignoring. Since this function doesn't allow returning error in signature then should panic if such an error occurs

pkg/auth/poptoken/shrpoptoken.go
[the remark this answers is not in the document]
wecha-ms-v2: as discussed, we can implement a simple nonce cache at the server side to check for repeated nonce ids. We won't implement the full specs.

pkg/auth/poptoken/nodeagentpoptokenscheme.go
wecha-ms-v2: sure I can make a more generic poptokenscheme and have our own scheme "inherit" it

pkg/auth/poptoken/nodeagentpoptokenscheme.go
```diff
@@ -0,0 +1,66 @@
...
+
+const (
+	tokenType  = "token_type"
+	reqCnf     = "req_cnf"
+	resourceId = "resourceid"
+)
+
+// Implements the interface for MSAL SDK to callback when creating the poptoken.
+// See AuthenticationScheme interface in https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/main/apps/internal/oauth/ops/authority/authority.go#L146
+type NodeAgentPopTokenAuthScheme struct {
```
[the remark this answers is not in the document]
wecha-ms-v2: I added the grpc object as "p" claim and the nonce claim

pkg/auth/poptoken/shrpoptoken.go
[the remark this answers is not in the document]
wecha-ms-v2: implemented simple nonce cache on server end.

pkg/auth/poptoken/msalauthprovider.go
wecha-ms-v2: the app cert is passed to the client in such a way that the cert/file will be refreshed automatically, i.e. it can change anytime while the app is running. 

longer run I can do something more fancy like refreshing the cert in the background, but opted for the easier solution first until we can finalize the integration.

pkg/auth/poptoken/rsakeymanager.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+func generatePrivateKey() (*rsa.PrivateKey, error) {
+	return rsa.GenerateKey(rand.Reader, RsaSize)
+}
+
+// Return a KeyPair. The keypair is its own copy and not a reference.
+func (r *rsaKeyManager) GetKeyPair(now time.Time) (*RsaKeyPair, error) {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
```
keystroke, line 42: I know my sample I shared had this, but thinking on it more we can drop this - it should be fine if we get multiple concurrent calls into this, as each returned instance gets a valid key for its lifetime and at worst we just have a couple rotations that could occur with last write wins

pkg/auth/poptoken/noncecache.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+		if isDelete {
+			delete(n.cache, nonce.Id)
+			n.queue = n.queue[1:]
+			n.size--
+		}
+	}
+}
+
+func (n *nonceCache) IsNonceExists(nonceId string, now time.Time) bool {
```
keystroke, line 63: the caller should pass a TTL instead, so we don't track insertion time but rather time that it expires. This way caller can set the proper duration for tracking based on the lifetime of the received token. Say its only valid for 3 more minutes, the caller would set it to expire in 3 mins. And if it was valid for 10 more minutes, we'd want to set the lifetime here to 10 minutes. as we discussed, pop tokens in the outer context only have a creation time, so the caller of this cache would have configuration for clock skew and "max pop token age" where it would reject a pop token whose created time was further into the past than "config.MaxPopAge + config.clockSkew". And the remaining time from this calculation would be the duration the nonce should be cached (maybe with some extra wiggle room).

pkg/auth/poptoken/shrpoptoken.go
keystroke, line 1: nit - this isn't "shr" specifically. can also consider dropping and just storing the headers and claims in map in the auth scheme type and only generating and signing the token when needed in format token call. this file itself is almost longer than my entire implementation

pkg/auth/poptoken/rsakeymanager.go
```diff
@@ -0,0 +1,80 @@
...
+		if err != nil {
+			return nil, err
+		}
+		r.privateKey = newPKey
+		r.createdDateTime = now
+	}
+
+	// Create and return a deep copy of the private key so clients are not impacted by a rotation midway.
+	privateKeyBytes := x509.MarshalPKCS1PrivateKey(r.privateKey)
+	privateKeyCopy, err := x509.ParsePKCS1PrivateKey(privateKeyBytes)
```
keystroke, line 55: if you do this inline as in my sample, the key can be private and unexported and you don't need deep copy as they are never muted and not exported

pkg/auth/poptoken/poptokenscheme.go
```diff
@@ -0,0 +1,59 @@
...
+	return a.shrPopToken.GenerateToken(accessToken, time.Now(), a.claims)
+}
+
+// Return the token type. Must be "pop"
+func (a *PopTokenAuthScheme) AccessTokenType() string {
+	return a.shrPopToken.Header.Typ
+}
+
+// Create a new instance of PopTokenAuthScheme. Pass in the custom claims to be set in the pop token here, e.g. resourceId
+func NewPopTokenAuthScheme(claims map[string]interface{}, rsaKeyPair *RsaKeyPair) (*PopTokenAuthScheme, error) {
```
keystroke, line 49: should not expose external management of raw key pair, hide this detail from caller unless we formalize a crypto provider interface that would implement strong key mgmt (which we won't do anyway). If you really want the key mgnr extra concept and file and test etc. just have this as an internal ref that caller does not init and interact with. Caller should never need to supply a raw key

pkg/auth/poptoken/raskeymanager_test.go
keystroke, line 1: typo in file name (i was wondering why the sorting was odd lol until i noticed that)

pkg/auth/poptoken/noncecache.go
wecha-ms-v2: I think you are thinking of the pop expiry? This is tracking the nonce cache expiry and not the pop token expiry. 
thiat is done in isTokenExpire(). [1] https://github.com/microsoft/moc/blob/user/wecha/pop2/pkg/auth/poptoken/nodeagentpoptokenvalidator.go#L82

keystroke: No I’m talking about for how long we should cache the nonces. Nonces never expire ever. If you ever see a repeat we’d want to reject. But, we obviously can’t remember all seen nonces forever. Instead, when we receive a nonce, we need to remember it for as long as the source context remains valid. That lets us know how long we need to cache it for to ensure the source token cannot be replayed. Once the source context is expired according to our time validation logic for token expiry, then it becomes “safe” to remove that nonce from our cache. Let the caller set the TTL is common pattern and clean abstraction.

pkg/auth/poptoken/poptokenscheme.go
```diff
@@ -0,0 +1,59 @@
...
+func (a *PopTokenAuthScheme) FormatAccessToken(accessToken string) (string, error) {
+	// append accessToken and our custom claims to the pop token.
+	// Note custom claims should be compatible with JWT specs, we don't expect these claims to be complex
+	return a.shrPopToken.GenerateToken(accessToken, time.Now(), a.claims)
+}
+
+// Return the token type. Must be "pop"
+func (a *PopTokenAuthScheme) AccessTokenType() string {
+	return a.shrPopToken.Header.Typ
+}
```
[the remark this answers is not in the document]
wecha-ms-v2: its meant to abstract the  wrapper from the actual pop token implementation.

pkg/auth/poptoken/rsakeymanager.go
wecha-ms-v2: will make it inline

madhanrm:
Auto-closing this PR as part of the monthly stale-PR cleanup - it has been open for over 90 days or has had no activity in the last 90 days. This is reversible: reactivate the PR if it is still needed. - automated cleanup on behalf of @madhanm

PR #427, opened 2026-04-10
Add vmProviderReady field to NodeInfo proto

niontive:
## Summary

Add `google.protobuf.BoolValue vmProviderReady = 10` to the `NodeInfo` message in `moc_common_nodeinfo.proto`.

## Problem

When wssdagent restarts, its VM provider runs a sequential bootstrap (~4.5 hours for ~800 VMs). During this window, the health endpoint reports the node as healthy because it has no readiness awareness. wssdcloudagent's `SelectAnyNode()` routes new VMs to the bootstrapping node, which rejects them with `ProviderNotReady`.

## Change

- Add `google.protobuf.BoolValue vmProviderReady = 10` to `NodeInfo` (field 10)
- Uses `BoolValue` (not bare `bool`) for backward compatibility:
  - `nil` = old agent (field not set) → treat as ready
  - `BoolValue{false}` = new agent, still bootstrapping → skip node
  - `BoolValue{true}` = new agent, ready → allow placement
- Regenerated `.pb.go` bindings via `make generate`

## Part of

Coordinated fix across 3 repos:
1. **moc** (this PR): Proto definition
2. **wssdagent**: Expose VM readiness via health endpoint
3. **wssdcloudagent**: Check readiness during placement

Bug 5: Node selection ignores VM provider readiness


go.mod
```diff
@@ -21,8 +21,8 @@ require (
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	golang.org/x/net v0.42.0 // indirect
 	golang.org/x/sys v0.34.0 // indirect
-	golang.org/x/text v0.33.0 // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20260120221211-b8f7ae30c516 // indirect
```
dhavalpopat, line 25: is this intentional to update the indirect dependencies?

dhavalpopat (dismissed):

go.mod
niontive: No, it was unintentional. I reverted go.mod/go.sum to match main and re-ran `go mod tidy` - no changes were needed. Fixed in the updated commit.

jocardeMSFT (approved):

dhavalpopat (approved):

PR #431, opened 2026-04-20
Add Azure Storage Blob as an image source

oukoedwin:
Adds a new AZURESTORAGEBLOB_SOURCE image source type to support pulling images from Azure Storage Blob thus extending the set of supported image sources beyond local, SFS, HTTP, and clone. This allows images to be pulled from Azure storage blob in AGC where `sfs` is not available

# Changes
- Added AZURESTORAGEBLOB_SOURCE = 4 to the ImageSource enum in moc_common_common.proto
- Regenerated the corresponding Go protobuf code (moc_common_common.pb.go)


rpc/common/moc_common_common.proto
```diff
@@ -392,10 +392,11 @@ extend google.protobuf.FieldOptions {
...
 enum ImageSource {
-	LOCAL_SOURCE = 0;
-	SFS_SOURCE   = 1;
-	HTTP_SOURCE  = 2;
-	CLONE_SOURCE = 3;
+	LOCAL_SOURCE             = 0;
+	SFS_SOURCE               = 1;
+	HTTP_SOURCE              = 2;
+	CLONE_SOURCE             = 3;
+	AZURESTORAGEBLOB_SOURCE  = 4;
```
raghavendra-nataraj, line 399: Where is the path to the source specified?

oukoedwin: The path to the source is JSON-marshalled into `sourcePath` in the [GalleryImage message ](https://github.com/oukoedwin/moc/blob/main/rpc/cloudagent/compute/galleryimage/moc_cloudagent_galleryimage.proto), consistent with how it is handled for the other download providers like SFS. 

The missing `*ImageProperties` RPC messages are used as JSON-unmarshall targets in `wssdagent`. In the current implementation, I create a fresh struct in wssdagent which is rather redundant so I should make an update in this direction. But they are functionally equivalent.

raghavendra-nataraj (dismissed):

oukoedwin:
Since this is the first PR for the AGC support, I have added a screenshot below of a successful ARB deployment (provisioned using the new `azurestorageblob` download provider). The underlying MOC makes use of binaries built from feature branches corresponding to these related PRs:

- [moc-sdk-for-go PR](https://github.com/microsoft/moc-sdk-for-go/pull/381/)
- [wssdagent PR](https://github.com/microsoft/wssdagent/pull/1522)
- [wssdcloudagent PR](https://github.com/microsoft/wssdcloudagent/pull/1702/)

[Here is the mini-design doc for the corresponding workstream](https://github.com/oukoedwin/wssdagent/blob/edwinouko/agc-image-provisioning/doc/support-azurestorageblob-download.md)

image

blairurish (dismissed):

rpc/cloudagent/compute/galleryimage/moc_cloudagent_galleryimage.proto
```diff
@@ -58,6 +58,15 @@ message AzureGalleryImageProperties {
...
 	string version = 2;
 }
+message AzureBlobImageProperties {
+	string catalogName = 1;
+	string audience = 2;
+	string version = 3;
+	string releaseName = 4;
+	uint32 parts = 5;
+	string cloud = 6;
+	string endpoint = 7;
```
rbalwani, line 68: On the nodeagent side,  VirtualHardDisk.source  is annotated  [(sensitivejson) = true] , and the JSON redactor ( pkg/redact/redact.go ) only redacts the keys  "private-key"  and  "sasURI" . The new  endpoint  field holds a blob URL — if a SAS-bearing URL is ever placed in it, the token would be logged unredacted, since  endpoint  isn't in the redaction key list.
Please confirm

rbalwani (dismissed):

rpc/cloudagent/compute/galleryimage/moc_cloudagent_galleryimage.proto
oukoedwin: The endpoint is just the storage account's FQDN so it does not contain SAS tokens or shared keys for any kind (the downstream Downloadsdk uses anonymous access). Regardless, I redacted the endpoint for future adaptability. Added a corresponding unit test as well as running manual tests for the incoming change

blairurish (approved):

rbalwani (approved):

